PR #13, opened 2020-06-14
chore: add support for `semantic-release`

soedirgo:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@supabase/pg-api",
-  "version": "0.0.9",
+  "version": "0.0.0",
```
kiwicopple, line 3: Is this correct @soedirgo ? Should it be back at 0?

kiwicopple:
Just one comment @soedirgo. Feel free to merge when you're happy !

package.json
soedirgo: Hmm, maybe `0.0.0-automated` is more explicit? The `version` field is unused since `semantic-release` infers the current version from the latest tag.

PR #21, opened 2020-06-21
refactor: Transition from JS to TS

soedirgo:
Mostly satisfying the type checker and changing import/exports. No change to logic or code structure. Also a small fix to the tests.


dist/server.js
```diff
@@ -1,29 +1,29 @@
...
-var express = require('express');
-var app = express();
+var express = require("express");
 var api = require('./api');
 var project = require('../package.json');
+var app = express();
 app.use(express.json());
 app.use(api);
-app.get('/', function (req, res) {
+app.get('/', function (_req, res) {
```
kiwicopple, line 9: good Elixir convention :)

soedirgo: Oh, it is? I think it was the TS language server yelling at me 😅

src/api/extensions.ts
```diff
@@ -13,4 +15,4 @@ router.get('/', async (req, res) => {
   }
 })
 
-module.exports = router
+export = router
```
kiwicopple, line 18: I've never seen this convention before. Should it be `export default router`? or is this a TS thing?

kiwicopple: ah - I think it's not a default export. Usually I'd do something like:

`export const router = Router()`

soedirgo: It's a TS thing yeah. Apparently it's the equivalent of CommonJS's `module.exports` object:
https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

dist/server.js
kiwicopple: it could be that too, i'm not very familiar with TS

kiwicopple:
OK looks great. I can't remember why I left the `dist` folder in but it makes PR's a bit annoying. I probably thought it would be easier for people to test (just run `node dist server.js`), but I think we should remove it now. Let's merge this then I'll strip it out and add to the gitignore

soedirgo:
👍 Good idea, was about to ask.

PR #22, opened 2020-06-22
refactor: Modularize src/lib/sql

soedirgo:
Extract SQL files in src/lib/sql into reusable subqueries.

Move complex queries (tables, roles) into the API side.


src/api/tables.ts
```diff
@@ -1,15 +1,91 @@
 import { Router } from 'express'
 
 import sql = require('../lib/sql')
-const { tables } = sql
+const { columns, grants, pk_list, relationships, tables } = sql
```
kiwicopple, line 4: Perhaps we should rename `pk_list` to `primary_keys` for consistency

src/api/tables.ts
```diff
@@ -1,15 +1,91 @@
...
+      relationships AS (${relationships})
+      SELECT
+        *,
+        COALESCE(
+          (
+            SELECT
+              array_to_json(array_agg(row_to_json(columns)))
+            FROM
+              (
+                SELECT
```
kiwicopple, line 26: See how this `COALESCE` is repeated for each subarray? I wonder if we can do a small refactor? something like

- add a file `./src/lib/helpers`
- add this function:

```js
export const coallesceRowToArray = (key, source, joinQuery) => {
  return `COALESCE(
          (
            SELECT
              array_to_json(array_agg(row_to_json(${source})))
            FROM
              ( ${joinQuery} ) ${source}
          ), '[]'
        ) AS ${key},`;
};

```
And usage:

```js
const sql = `
      WITH tables AS (${tables}),
      columns AS (${columns}),
      grants AS (${grants}),
      pk_list AS (${pk_list}),
      relationships AS (${relationships})
      SELECT
        *,
        ${coalesceRowToArray(primary_keys, pk_list, 'SELECT * FROM pk_list WHERE pk_list.table_id = tables.table_id')}
    //....
    `
```
I think it will be cleaner (but feel free to improve this function if it seems confusing)

src/api/tables.ts
soedirgo: 👍 Agreed

src/api/tables.ts
soedirgo: Good point, I just glanced over these without noticing they look similar 😅.
What about combining `key` and `source`? Seems like they're all the same (after renaming `pk_list` to `primary_keys`).

kiwicopple: note - just fixed some typos

kiwicopple: yeah you could actually just combine them. I can't think of a time that we would want to rename the `source` key, but if we do, then the dev can just type it out in full

Issue #17, opened 2020-06-19
Ability to create a new role


kiwicopple:
A user should be able to post to `/roles` which should create a new role in the database

PR #24, opened 2020-06-23
feature: Create roles and tables from endpoints

### LINKED ISSUE #16: Ability to create a table
A user should be able to post to `/tables`, which should create a new table in the database

soedirgo:
Sticking with plain string for `Tables.Column['table']`, circular references make fetching awkward (fetch table `T` which has column `C`, fetch column `C` which has table `T`, fetch table `T` which has column `C`, ...)

There are many more configurations we can make on creation (especially for tables), but we'll stick with the basics for now.

I haven't tested them much (just sanity test that they actually create roles/tables), fingers crossed they don't break something...

Closes #16, #17.


kiwicopple:
Everything seems great, but I think no tests right? Could you throw in a couple of tests for the new POST endpoints?

soedirgo:
One weird thing I found is when I add a new role (just raw SQL), the the result from `/roles` doesn't seem to have updated somehow (though it's updated in DBeaver). Maybe I'm missing something obvious?

Otherwise I'll just add the test for `roles` later and track the issue down tomorrow.

kiwicopple:
Yeah I think it's because the `/roles` endpoint is faulty - it's not returning all the roles: https://github.com/supabase/infrastructure/issues/14

soedirgo:
Roles should work OK now. Will merge this and close [supabase/infrastructure#14](https://github.com/supabase/infrastructure/issues/14) if this looks good.

kiwicopple:
amazing thanks Bobbie, looks like you did some nice tidying up. Once you've merged this, let's also push to `master` to bump the deployment. I'll try get this version out into production this release cycle

PR #27, opened 2020-06-30
refactor: update ID fields for the types

soedirgo:
I tried maintaining the current fields, so the SQL templates look rather messy right now (a lot of extra `INNER JOIN`s that make queries slow). Will do some polishing before the next release.

Drafting this until Wed.


src/lib/sql/columns.sql
```diff
@@ -1,30 +1,22 @@
+-- We can uniquely point to a specific column using table_id + ordinal_position.
```
kiwicopple, line 1: This isn't included in this PR right?

soedirgo: There's no explicit ID field for it. Do you want it to be separate field? (e.g. `${table_id}.${ordinal_position}`)

kiwicopple: Yeah I think something like that makes sense - saves the user from creating a composite key each time they want to update a column

PR #36, opened 2020-07-21
docs(openapi): add initial OpenAPI spec

soedirgo:
This adds the initial OpenAPI spec for pg-api.

Most of the skeleton is in place, but it's still missing some things, e.g.:

- Non-boiletplate examples
- Failing HTTP responses
- Descriptions
- Test against an OpenAPI validator
- Verify required/nullable fields against actual properties on Postgres

I don't think we can get this right the first time, so let's go with this and Kaizen the rest.

Also, this spec slightly differs from the current implementation (camelCase vs. snake_case, type inconsistencies). These will be addressed in the future.

P.S. To view the docs, run `npm run dev:watch` in the `docs` directory.

### DIFF
--- a/docs/api.yml
+++ b/docs/api.yml
@@ -1,0 +1,1114 @@
+openapi: 3
+
+info:
+  title: Postgres API
+  description: |
+    Manage your PostgreSQL database using a RESTful API.
+
+    This is still in early development, so most of the functionality is read-only.
+
+    The goal of this API is to enable the full management of a Postgres database using a RESTful interface. This includes adding tables, columns, roles, and users at runtime, as well as running ad-hoc queries.
+
+    Made by Supabase: https://supabase.io
+
+    # Getting started
+
+    ## Usage
+
+    > Basic usage
+
+    ```sh
+    curl -X GET http://localhost:1337/ \
+    -H 'Content-Type: application/json' \
+    -H 'X-Connection-Encrypted: ENCRYPTED_STRING'
+    ```
+    ```js
+    const data = await fetch('http://localhost:1337', {
+    method: 'GET',
+    headers: {
+        'Content-Type': 'application/json',
+        'X-Connection-Encrypted': 'ENCRYPTED_STRING'
+    }
+    })
+    ```
+
+    For security reasons, this API is best self-hosted and set up with ENV_VARS with the default connection string.
+
+    If you want to use this with multiple Postgres instances, you can pass the connection string as a header but it must be encrypted.
+
+
+    ## Self Hosting
+
+    ```
+    https://github.com/supabase/pg-api
+    ```
+
+    We support several different methods for self-hosting, detailed in the [repository](https://github.com/supabase/pg-api).
+
+
+    ## Authentication
+
+    Authentication is not provided. Make sure you use this inside a secure network or behind a secure API proxy.
+  contact:
+    name: Support
+    url: https://github.com/supabase/pg-api/issues/new/choose
+    email: support@supabase.io
+  license:
+    name: Apache 2.0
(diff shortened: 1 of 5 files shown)


kiwicopple:
This seems great @soedirgo ! I wish we could generate the Typescript interfaces from the API spec too, then the spec becomes the source of truth for docs, testing, and type generation. 

I know you had some hesitations. What are your thoughts so far? Is this a feasible solution long-term?

soedirgo:
I'm in favor of keeping it, if only to have one source of truth, but as I said I can't get this right the first time, so I'll have to incrementally improve it.

On that note, do we still need the interfaces once we have this? AFAIK the interfaces only apply for `GET` results (not request body), though ideally we want to generate this kind of boilerplate for the server side and plug our implementation in (I'll look into this).

kiwicopple:
> I'm in favor of keeping it

Nice, let's keep then.

> do we still need the interfaces once we have this? 

They do provide some level of type safety (eg, [here](https://github.com/supabase/pg-api/blob/7223a88a31e59d6425881180120c6764ee736445/src/api/schemas.ts#L129)), but it would be more useful if we just had a single source of truth. If we can generate them down the road using the swagger, that would be a massive win, if you feel like they're a burden right now then let's drop them

soedirgo:
I'll see if I can generate it automatically. My gripe with the interfaces is that TypeScript doesn't provide any type-checking on runtime, so e.g. if you declare a variable as `boolean`, and you give it a value in a way that the type system can't check (like HTTP request body), the compiler doesn't complain.

I've only been using the interfaces as the single source of truth, but have the OpenAPI spec for that now.

kiwicopple:
> My gripe with the interfaces is that TypeScript doesn't provide any type-checking on runtime

Agreed, seems a bit pointless. Perhaps we can tie it in with a data validator of some sort? (eg: something like this https://www.npmjs.com/package/validate-typescript)

kiwicopple:
Another thought Bobbie: Ant seems to be going all-in on Go after using gotrue. `pg-api` is still a very small API, with most of the functionality in the SQL code. It's not too late to switch to a strongly typed language (go/rust). Let's discuss on monday at the breakfast.

There are some downsides to moving away from Nodejs 

- node/js has the largest community, which makes it possible for people to read/contribute to the repo. In saying that, we're the only ones contributing right now so it should be a small factor when considering the right tool for this job. 
- we are supporting many languages which could decrease our velocity (elixir-realtime, go-gotrue, frontend-js/node, pgrst-haskell).

soedirgo:
> It's not too late to switch to a strongly typed language (go/rust).

I've been contemplating switching to Go for quite a while :). Seems like the perfect use case for Go (server-side/cloud infra stuff).

kiwicopple:
Hey @soedirgo, Ant has gone for Go on the admin API. If you want a project, feel free to switch this over too - I'll leave it up to you. I won't be coding much for the next ~2 weeks so feel free to take lead on this

soedirgo:
Merging this for now, will continue the progress in a separate branch.

kiwicopple:
Hey @soedirgo did you start a different branch? I'm will try jump into a few of the issues for Joshen later this week. 

Is there a new workflow now - I imagine I can just update the OpenAPI spec, add the code, then generate the docs?

soedirgo:
> Hey @soedirgo did you start a different branch?

I haven't, prioritizing other things at the moment

> I imagine I can just update the OpenAPI spec, add the code, then generate the docs?

Yeah, unfortunately you'll have to generate the docs manually for now. Will work on automating this to you only need to update the OpenAPI spec.

kiwicopple:
OK no probs - can do it manually for now. Just wanted to check!

PR #48, opened 2020-08-07
fix: don't run name-change SQL when unchanged

### LINKED ISSUE #44: Don't run name updates on `/tables` and `/columns` if the name is not changed
This line:

https://github.com/supabase/pg-api/blob/f7d66c2395e682d4ebb082fe881fe1662052b5e2/src/api/tables.ts#L87

Should only run if the name is different from what the table was previously named. 

At the moment it is running even if you pass it the same name. This results in an error `relation already exists`.

soedirgo:
Covers /tables & /columns on PATCH.

Closes #44.

### DIFF
--- a/src/api/columns.ts
+++ b/src/api/columns.ts
@@ -168,7 +168,7 @@ const alterColumnSqlize = ({
   comment?: string
 }) => {
   const nameSql =
-    name === undefined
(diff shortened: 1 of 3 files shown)


kiwicopple:
amazing thanks bobbie

PR #60, opened 2020-09-18
Adds a test for errors

### LINKED ISSUE #55: Return meaningful error messages from API
Context: Whenever the editor faces an error from calls to pg-api, I'll just show the error given by the API as a toast message. 

At the moment, i noticed that most of the messages just throw "Request failed with status code 500", along with a 500 status code - will be better if the API returns more specific messages for greater clarity in UX and debugging

kiwicopple:

Creates test for errors

- Fixes #55

### DIFF
--- a/src/api/policies.ts
+++ b/src/api/policies.ts
@@ -48,7 +48,7 @@ router.post('/', async (req, res) => {
   } catch (error) {
     // For this one, we always want to give back the error to the customer
     console.log('POST error!', error)
-    res.status(200).json({ error: error.toString(), status: 200 })
+    res.status(400).json({ error: error.toString() })
   }
 })
 
@@ -85,7 +85,7 @@ router.patch('/:id', async (req, res) => {
   } catch (error) {
     // For this one, we always want to give back the error to the customer
     console.log('Soft error!', error)
-    res.status(200).json({ error: error.toString(), status: 200 })
+    res.status(400).json({ error: error.toString() })
   }
 })
 
--- a/src/api/policies.ts
+++ b/src/api/policies.ts
@@ -48,7 +48,7 @@ router.post('/', async (req, res) => {
   } catch (error) {
     // For this one, we always want to give back the error to the customer
     console.log('POST error!', error)
-    res.status(400).json({ error: error.toString() })
+    res.status(400).json({ error: error.message })
   }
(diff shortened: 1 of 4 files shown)


kiwicopple:
Hey Joshen, Bobbie -> we are actually returning soft errors, perhaps not in the correct format though?

I've added a test which shows how it is returned. It has a rather annoying `error:` prepended to the string, which seems to come from `node-postgres`

soedirgo:
I'm not really clear about which kind of errors should be soft errors (and vice versa). I think errors coming from the SQL editor should be "soft" (show error in query result, not toast), but maybe this can be done without soft errors (i.e. 200 OK but error response body) by just catching the error on the SQL editor side. I prefer this because `/query` can also be seen as an escape hatch for more complicated DDL whose error we don't want to show to the user.

Errors like invalid type should show a toast, and 5xx errors should show... a big toast (once in a blue moon event)? But again, this can be handled without 200 response with error body.

kiwicopple:
Good call, after thinking more returning a 200 doesn't really make sense. Let's make it a [400 bad request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400)

What I'm thinking:

```js

res.status(400).json({ error: error.toString() })

```

returns 

```js

{
  status: 400, // bad request
  data: {
    error: 'PG error message'
  }
}
```

kiwicopple:
Will leave you to review @soedirgo - by changing to a 400 we may actually end up breaking the dashboard but I think it's technically more correct to return hard errors from the server

kiwicopple:
FYI, `res.status(400).json({ error: error.toString() })` should probably be `res.status(400).json({ error: error.message || error.toString() })`

soedirgo:
👍 Using `error.message` should be enough, since this is a [required property](https://nodejs.org/docs/latest-v12.x/api/errors.html#errors_error_message) for Error objects in Node.

kiwicopple:
@soedirgo @phamhieu @joshenlim FYI I am going to merge this in. It may be a breaking change as I am now returning `400` rather than `200`. I just want to give you a heads up - in case you see strange behaviour on the front end. The Semantic release will bump the version, so it will proabbly appear in `pg-api` version `0.10.5`

PR #64, opened 2020-09-30
fix: pkg missing a dynamic require in pg-format

soedirgo:

Bug fix.

pkg is missing [a dynamic require](https://github.com/datalanche/node-pg-format/blob/3157b520bfec69b05c20f578537c0fe8861f6edf/lib/index.js#L4), so the resulting binaries fail to run.

## What is the new behavior?

Add the require to `pkg.scripts` in `package.json`, as per [docs](https://github.com/vercel/pkg#config).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,9 +91,9 @@ jobs:
         run: |
           npm ci
           npm run dist
-          tar -czvf pg-api-linux.tar.gz -C ./bin start-linux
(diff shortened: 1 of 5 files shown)


soedirgo:
I also removed the binaries as we have releases now, but feel free to add back if releases won't do.

kiwicopple (approved):
> removed the binaries

Nice one 👍

Issue #65, opened 2020-10-01
Manage subscriptions


kiwicopple:
# Feature request

## Is your feature request related to a problem? Please describe.

The goal is to be able to manage subscriptions for 

## Describe the solution you'd like

- List subscriptions
- Enable replication for all tables
- Disable replication for all tables
- Enable for individual table (only if it's not "for all tables")
- Disable for individual table (only if it's not "for all tables")
- Enable replication type "FULL" for each table

Possible:

- Add replication status for each table to the `/tables` route

## Additional context

- Related (internal task) https://github.com/supabase/infrastructure/issues/193

soedirgo:
Would making a new endpoint (`/publications`) be a good idea? Though realistically, we're only using `/pg/publications/supabase_realtime`.

kiwicopple:
Yes I think it's a good idea. Even though we only use one, we could create several publication in the future (for replication)

PR #67, opened 2020-10-10
Add support for /publications

soedirgo:

Feature.

## What is the new behavior?

Support managing publications through the `/publications` endpoint.

## Additional context

Progresses on #65, setting `REPLICA IDENTITY FULL` will be done through `/tables`.


kiwicopple:
Awesome work here Bobbie - very nice

Issue #61, opened 2020-09-20
Unable to set defaultValue for column text type


joshenlim:
# Bug report

## Describe the bug

Attempting to set a defaultValue for a column of text type returns a 500

Doing it so for int type columns seem to be working fine

## To Reproduce

Via the table editor:

1. Create a new column
2. Set column type to text and fill in any default value
3. Select save
4. A toast error will show 500 that's returning from pg-api

## Expected behavior

Column should save with provided defaultValue

## Screenshots

![image](https://user-images.githubusercontent.com/19742402/93708949-01e4cb80-fb6d-11ea-95d2-5012da5c2679.png)

![image](https://user-images.githubusercontent.com/19742402/93708950-0a3d0680-fb6d-11ea-9377-ba431f5412a1.png)

## System information

- OS: macOS
- Browser (if applies): Chrome
- Version of supabase-js: 0.35.11
- Version of Node.js: 11.9.0

PR #68, opened 2020-10-10
Allow `default_value` to be literals or expressions

soedirgo:
Literals: `'hello'`, `123`

Expressions: `NULL`, `NOW()` (function calls)

Bug fix.

Can't use expressions as default values.

## What is the new behavior?

Can use expressions by setting the `default_value_format` field.

## Additional context

Related: #61, #62.

@kiwicopple @joshenlim I'm deviating a bit from how we discussed it - instead I'm just adding a field so the body format is kept flat, and not break existing code (except for e.g. `NULL` and `NOW()`).


kiwicopple:
> instead I'm just adding a field so the body format is kept flat, and not break existing code

Good one @soedirgo, I think that's a good idea . @joshenlim if you're happy let's merge

joshenlim:
Yup this looks good to me! :) Will add a ticket as well on the editor to update accordingly too. Skimming through the changes - so default_value_format is assumed to be 'literal' by default right?

soedirgo:
Yep, so no need to add quotes unless doing e.g. `NOW()`.

PR #73, opened 2020-10-25
Rename row count estimates

### LINKED ISSUE #66: live_count isn't always accurate
# Bug report

## Describe the bug

The live_count on tables is only indicative, but we are using it on the frontend to determine the number of rows. We should see if there is a figure that is accurate always

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. insert thousands of rows into a database
2. query `/tables`
3. check `live_count`

## Screenshots

image

Don't worry about the error he mentions, but this is the Dashboard after he has inserted 20m rows:

image

soedirgo:

Bug fix.

## What is the new behavior?

Rename row count estimates from the statistics collector.

## Additional context

Closes #66.

As I mentioned in #66, we should be explicit that these numbers are estimates.

P.S. this doesn't exactly solve the problem in #66, that one needs either exact count (bad for large tables) or `ANALYZE` (but it's [automatically run](https://www.postgresql.org/docs/12/runtime-config-autovacuum.html#GUC-AUTOVACUUM-ANALYZE-THRESHOLD) for large inserts/updates/deletes by default anyway--but the user will have to wait). So... wontfix?

### DIFF
--- a/src/lib/sql/tables.sql
+++ b/src/lib/sql/tables.sql
@@ -14,8 +14,8 @@ SELECT
   pg_size_pretty(
     pg_total_relation_size(format('%I.%I', nc.nspname, c.relname))
   ) AS size,
-  pg_stat_get_live_tuples(c.oid) AS live_row_count,
-  pg_stat_get_dead_tuples(c.oid) AS dead_row_count,
+  pg_stat_get_live_tuples(c.oid) AS live_rows_estimate,
+  pg_stat_get_dead_tuples(c.oid) AS dead_rows_estimate,
   obj_description(c.oid) AS comment
 FROM
   pg_namespace nc


kiwicopple:
I wonder if this is going to have a significant impact on performance especially since it needs to fetch the estimated count, then potentially fetch the actual count. Let's wait and see I guess - it seems like this is important enough to warrant it.

> So... wontfix?

Agreed

> Should I update the name to live_row_count_estimate (breaking change)?

Yeah, I think we should rename (don't write breaking change in the git commit though - no need to bump the major version)

soedirgo:
> I wonder if this is going to have a significant impact on performance

I did a little check on this. As for getting the estimate, it's the same one as what we're currently showing. The exact count is only executed if the condition (n < 2000) is true.

soedirgo:
@kiwicopple Any follow up for this? I'm reconsidering just leaving the estimate logic be and just renaming the field. If FE needs something more sophisticated we can revisit.

kiwicopple:
> leaving the estimate logic be and just renaming the field

That makes sense - make it explicit. Go for it Bobbie

PR #74, opened 2020-10-25
Support setting identity_generation in /columns

### LINKED ISSUE #69: Allow Column CRUD to specify {ALWAYS | BY DEFAULT} for identity columns
# Feature request

When creating or updating columns, we can specify a column to be an identity column by passing `isIdentity: false/true` to the POST payload body.

We'll need an added level of detail to isIdentity by allowing users to specify if it should be `GENERATED ALWAYS` or `GENERATED BY DEFAULT` as [such](https://www.postgresqltutorial.com/postgresql-identity-column/#:~:text=The%20GENERATED%20ALWAYS%20instructs%20PostgreSQL,value%20for%20the%20identity%20column.), as they both have different implications on whether the user can specify a value for the column.

`GENERATED ALWAYS` prevents user from specifying a value for that column, but `GENERATED BY DEFAULT` allows so. In both cases, if no value is specified, the value will be generated automatically in a sequence.

Currently, the FE table editor does not allow user to specify a value if the column is marked as `isIdentity`, but this behavior will need to be changed accordingly to above.

Add an additional parameter for the POST payload body to allow the specification of identity format as above.
And also return the parameter when getting columns.

soedirgo:

Feature.

## What is the new behavior?

Allow setting `identity_generation` for identity columns as either `ALWAYS` or `BY DEFAULT`, like so:
```js
await axios.post(`.../columns`, {
  table_id: ...,
  name: 'foo',
  type: 'int',
  is_identity: true,
  identity_generation: 'ALWAYS'
})

await axios.patch(`.../columns/10000.1`, {
  is_identity: true,
  identity_generation: 'BY DEFAULT'
})
```
Note that the column has to be `is_nullable: false` for it to be an identity column.

## Additional context

Closes #69. cc @joshenlim


kiwicopple:
Looks awesome bobbie. OK to merge?

PR #75, opened 2020-11-13
fix: propagate errors from Postgres

soedirgo:

Bug fix.

Most error responses are `500 Server Error` with error message `Database error.`, which is not very helpful.

## What is the new behavior?

Error handling goes like this:
- Baseline response on error is 400 Bad Request, propagating Postgres's error message
- If the request is infallible (e.g. GET with no payload/query params), respond with 500 Server Error, again propagating Postgres's error message
- More specific responses can be done as more specific errors are found (404 for missing IDs, etc.)


kiwicopple:
This is much better. thanks Bobbie

PR #78, opened 2020-12-14
perf: Inline SQL information_schema joins

### LINKED ISSUE #43: perf: inline all SQL on information_schema joins
Based on @soedirgo's research here: https://github.com/supabase/pg-api/issues/41

we are `JOIN`ing `information_schema` views with system tables, e.g.:

```sql
-- tables.sql
...
FROM
  information_schema.tables
  JOIN pg_class c ON quote_ident(table_schema)::regnamespace = c.relnamespace
  AND c.relname = table_name
  LEFT JOIN pg_stat_user_tables ON pg_stat_user_tables.schemaname = tables.table_schema
  AND pg_stat_user_tables.relname = tables.table_name
...
```

`information_schema` views already use system tables under the hood (like below), so we could inline it, though it'd make the SQL templates pretty ugly.

```sql
-- information_schema.tables
...
FROM pg_namespace nc
  JOIN pg_class c ON nc.oid = c.relnamespace
  LEFT JOIN (pg_type t
  JOIN pg_namespace nt ON t.typnamespace = nt.oid) ON c.reloftype = t.oid
...
```

The performance difference is big though (10x):

```
# Not inlined
❯ hyperfine 'curl -s http://localhost:1337/tables' --warmup 5
Benchmark #1: curl -s http://localhost:1337/tables
  Time (mean ± σ):      1.004 s ±  0.027 s    [User: 4.3 ms, System: 5.5 ms]
  Range (min … max):    0.983 s …  1.063 s    10 runs

# Inlined
❯ hyperfine 'curl -s http://localhost:1337/tables' --warmup 5
Benchmark #1: curl -s http://localhost:1337/tables
  Time (mean ± σ):      97.6 ms ±   5.3 ms    [User: 4.2 ms, System: 5.1 ms]
  Range (min … max):    91.9 ms … 110.0 ms    27 runs
```

soedirgo:

Perf improvement.

Some of the SQL templates we use do joins on `information_schema` views.

## What is the new behavior?

Do joins on system catalogs instead (which the views use under the hood).

Also removes some fields we don't use.

## Additional context

Closes #43.

The PR is scant on details here, let me know if any change feels out of place.


src/api/tables.ts
```diff
@@ -151,8 +151,8 @@ const getTablesSql = (sqlTemplates) => {
       FROM
         relationships
       WHERE
-        (relationships.source_schema = tables.schema AND relationships.source_table_name = tables.name)
-        OR (relationships.target_table_schema = tables.schema AND relationships.target_table_name = tables.name)`
+        (relationships.source_schema :: text = tables.schema AND relationships.source_table_name :: text = tables.name)
+        OR (relationships.target_table_schema :: text = tables.schema AND relationships.target_table_name :: text = tables.name)`
```
soedirgo, line 155: `source_schema` is `regnamespace`, need to convert to do comparison. Same for others.

src/lib/sql/columns.sql
```diff
@@ -1,32 +1,86 @@
...
   c.oid :: int8 AS table_id,
-  table_schema AS schema,
-  table_name AS table,
-  (c.oid || '.' || ordinal_position) AS id,
-  ordinal_position,
-  column_name AS name,
-  column_default AS default_value,
-  data_type,
-  udt_name AS format,
-  col_description(c.oid, ordinal_position) AS description,
```
soedirgo, line 11: Use `comment` instead.

src/lib/sql/columns.sql
```diff
@@ -1,32 +1,86 @@
...
-  is_identity :: boolean,
-  identity_generation,
-  is_nullable :: boolean,
-  is_updatable :: boolean,
```
soedirgo, line 15: `BASE TABLE` is always updatable.

src/lib/sql/schemas.sql
```diff
@@ -1,12 +1,15 @@
...
+
 SELECT
-  nsp.oid AS id,
-  catalog_name,
-  schema_name AS name,
-  schema_owner AS owner,
-  default_character_set_catalog,
-  default_character_set_schema,
-  default_character_set_name,
-  sql_path
```
soedirgo, line 9: Unsupported in Postgres ([ref](https://www.postgresql.org/docs/13/infoschema-schemata.html))

src/lib/sql/tables.sql
```diff
@@ -1,46 +1,33 @@
...
-  information_schema.tables
-  JOIN pg_class c ON quote_ident(table_schema)::regnamespace = c.relnamespace
-  AND c.relname = table_name
-  LEFT JOIN pg_stat_user_tables ON pg_stat_user_tables.schemaname = tables.table_schema
-  AND pg_stat_user_tables.relname = tables.table_name
+  pg_namespace nc
+  JOIN pg_class c ON nc.oid = c.relnamespace
 WHERE
-  table_type = 'BASE TABLE'
+  c.relkind IN ('r', 'p')
```
soedirgo, line 24: `BASE TABLE` is `r` (normal) or `p` (partitioned)

kiwicopple:
Awesome Bobbie - I don't have great wifi in the airport so I can't test myself. Did it speed up the query? If there are no breaking changes, let's get this one in 🚢

soedirgo:
Oof. Not seeing the expected 10x in #43, your `MATERIALIZED` commit already brought it down to ~200ms from ~1s. Now it's ~150ms with `MATERIALIZED` and ~90ms without. Not as impressive...

No intended breaking changes here for the FE, so here's hoping there's no *unintended* breaking changes 😁

kiwicopple:
That's fine, it's still a massive improvement, and the `MATERIALIZED` only works for PG 11+ I think.

Good job @soedirgo - feel free to merge whenever you want

PR #86, opened 2020-12-31
feat: improve logging with pino

### LINKED ISSUE #83: Improve logging
## Goals
- Our pg-api middleware handles requests for multiple users and it is hard to filter logs per project. We should be able to easily filter logs per project
- Filter based on log levels like critical, error, info, etc
- printing logs in JSON makes it easier for searching in Cloudwatch. Right now each new line is treated as a separate log message which makes it harder to group and search. 
- log information about the incoming query, IP address of client, etc - useful for debugging issues and replaying the exact request which caused the error
This is how the logs appear in Cloudwatch currently for context
image

Lets move to using structured logging with something like [Winston](https://www.npmjs.com/package/winston). We can add DB_HOST as labels, add logging levels, timestamps.

soedirgo:

Feature.

Logging is done with `console.log`s

## What is the new behavior?

Improve logging with [pino](https://github.com/pinojs/pino). Among other things that get logged:
- Log levels (`info`: `30`, `error`: `50`)
- Postgres connection string
- Request body

## Additional context

Closes #83.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "express": "^4.17.1",
         "pg": "^7.0.0",
         "pg-format": "^1.0.4",
+        "pino": "^6.8.0",
         "sql-template-strings": "^2.2.2"
       },
       "bin": {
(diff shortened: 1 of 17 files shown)


kiwicopple:
Nice one. We could even send some of these logs through LogFlare now: 

https://github.com/Logflare/pino-logflare

PR #87, opened 2021-01-09
Separate API and express server

kiwicopple:

This is v1 of the separartion between the express server and the code. The motivation is that we can use this in our CLI (for dumping schemas to disk/repos.

## Additional context

This is a very small start to verify that we're on the same wavelength.

I think in the next PR i should refactor the `connectionString` param out of each function, which is probably more functional than we require for this lib. 
Instead we can follow the pattern we used for our other libraries to init a client which holds the connection info. eg:

```js
import { createClient } from '@supabase/pg-api'

const pg = createClient(connectionString)

const publicSchema = pg.schemas.byName('public')
```


kiwicopple:
@soedirgo this was about all I could do on the plane (also I wanted to keep the PR small/readable). 
Let me know if you think it's in the right direction. If you like, merge and then I'll continue refactoring on a new PR

soedirgo:
> Instead we can follow the pattern we used for our other libraries to init a client which holds the connection info.

Agreed with this , but how about calling `new PostgresApi(...)` instead of `createClient(...)`? Stripe does [something similar](https://stripe.com/docs/js/initializing). This is something that's been bugging me in supabase-js as well, since I don't see the pattern often.

I was thinking of separating the Express code with the routes completely, so we'd have a library `@supabase/postgres-api` that exports `schema`, `table`, `column`, etc. so either the server or the CLI doesn't interact with `RunQuery` at all.

Some questions on the peripheral:
- Should we keep `include_system_schemas`? Since we're trying to abstract away DDL here, I think it's safe to omit this feature, otherwise there's always `/query`
- Right now connections are pretty much unpooled (since the pool is closed at the end of the query). Should we start pooling the connections? This would mean changing a DB config would require either a restart or setting the config for the session as well.

Sorry if my thoughts are a bit scattered here, there are some things in my mind that I can't find the right time to bring up 😅

kiwicopple:
> but how about calling new PostgresApi(...) 

Good one - more consistent with our sub-libraries. The reason we have `createClient` in our supabase-js lib is so that we could keep some parts modular (rest, admin, emails, maps, etc), and then the `createClient` would be a convenience function for instantiating them all together or leaving some parts out. There is now a one-to-one mapping between `SupabaseClient` and `createClient`, but that wasn't the intention at the start 

> Should we keep include_system_schemas? Since we're trying to abstract away DDL here, I think it's safe to omit this feature

I don't know why we wouldn't keep it? It seems useful to include system info, especially when we want to fetch `types`. I don't think our goal is really to abstract away the DDL - it's to give Postgres an easier (familiar) DX.

> Right now connections are pretty much unpooled (since the pool is closed at the end of the query). Should we start pooling the connections? This would mean changing a DB config would require either a restart or setting the config for the session as well

This is 100% a good idea. Should we start an issue for this?

soedirgo:
> especially when we want to fetch types.

Ahh yeah, that'd be an obvious use case. Agreed with keeping it in for now.

> This is 100% a good idea. Should we start an issue for this?

I think this is just a few lines of diff, so can be done along with the split.

kiwicopple:
> Any thoughts on separating all the node-postgres stuff from the server + CLI? It makes for a cleaner boundary IMO.

Can you give an example of what you mean here?

Also, are we good to merge this one? I will then continue improving in a new PR because I don't want this one to become a behemoth

kiwicopple:
Actually it looks like the tests are failing for some reason. Odd - they were fine on my local dev

soedirgo:
Let me make a quick skeleton in a bit, a repo is worth a googol words ;)

kiwicopple:
OK tests are passing. Ready to merge whenever you're happy @soedirgo 👍

soedirgo (approved):

Issue #89, opened 2021-01-28
RFC: Renaming this repo


kiwicopple:
We would like to rename this repo to make it clearer on first glance exactly what this is doing. `pg-api` seems like it could be for interacting with data (like PostgREST), but the intent of the repo is to manage the database itself.

Our current proposal is `postgres-meta`.

kiwicopple:
Renamed and relabeled. Remaining:

- [ ] update all release/build artifacts

PR #91, opened 2021-02-03
feat: split up the API and the server

### LINKED ISSUE #71: Proposal/RFC: split up the API and the server
At the moment all of the interactions work through an express server. However, when we start work on the CLI, we will need to use this functionality without express. My proposal is to split the API and Server functioanality into 2 areas

```bash
src
- api      # contains all the TypeScript classes
- server   # contains the express routes, which import from `../api`
```

Then, we should be able to do something this:

```sh
npm install @supabasee/pg-api
```

### API (usage with our CLI)

```js
import { createClient } from 'pg-api'

const api: PostgresApiClient = createClient(process.env.connection_string)

api.tables.add({
  schema: 'public',
  name: 'test'
})
```

### Server

```js
import { createServer } from  'pg-api'

const server: PostgresApiServer = createServer(process.env.connection_string, {})

server.start() // starts express server on port 3000
```

Then the route `POST /tables` calls `api.tables.add()`

soedirgo:

Feature.

The API logic and the server are tied up.

## What is the new behavior?

Split them up into:
- an npm package: @supabase/postgres-meta,
- a server runnable with `npm start` or the packaged executables.

## Additional context

Closes #71, #89.

Commits are organized by commits for ease of review.

What this PR does:
- add the library files to `src/lib`,
- move all server-related things (Express, logging, etc.) to `src/server`,
- old routes are left in `src/api` and some middleware are left in `src/lib`,
- minor changes in the config files to publish to npm,
- finalize renaming `pg-api` to `postgres-meta`.

In the future, the only things to keep are `src/lib` (minus `connectionMiddleware.ts` & `connectionPool.ts`) and `src/server`.

Things renamed (@inian FYI):
- Repo name & url, docs, names in config files
- Docker image name
- Binary name (`pg-api-<os>` -> `postgres-meta-<os>`)
- Envs (`PG_API_*` -> `PG_META_*`)

@kiwicopple for the server, I didn't change the existing routes at the moment, since I'm moving enough things around already to be anxious of something breaking. But using `src/server/routes/schemas.ts` passes all tests.


src/server/routes/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import CryptoJS from 'crypto-js'
+import { Router } from 'express'
+import { PG_CONNECTION, CRYPTO_KEY } from '../constants'
+import logger from '../logger'
+import oldRoutes from '../../api'
+
+/**
+ * Adds a "pg" object to the request if it doesn't exist
+ */
+const addConnectionToRequest = async (req: any, res: any, next: any) => {
```
w3b6x9, line 11: @soedirgo isn't this the same `addConnectionToRequest` as the one being exported from `src/lib/connectionMiddleware.ts`?

w3b6x9: @soedirgo 

> minus connectionMiddleware.ts

Oh, I see!

src/lib/PostgresMeta.ts
```diff
@@ -0,0 +1,43 @@
...
+import PostgresMetaPublication from './PostgresMetaPublication'
+import PostgresMetaRole from './PostgresMetaRole'
+import PostgresMetaSchema from './PostgresMetaSchema'
+import PostgresMetaTable from './PostgresMetaTable'
+import PostgresMetaType from './PostgresMetaType'
+import PostgresMetaVersion from './PostgresMetaVersion'
+import { init } from './db'
+
+export default class PostgresMeta {
+  query: Function
```
w3b6x9, line 16: @soedirgo I see `Function` type used for `query` in a couple of places so I wanted to suggest an alternative.

Perhaps something like this is more descriptive, we can use it in all the other places `query` comes up, and help us down the line as we continue development:

```
type queryAsyncFunc = {
  (sql: string): Promise<{ data: any[] | null, error: { message: string } | null}>;
}

query: queryAsyncFunc
```

soedirgo: Yeah, the typings aren't final (there are still some `any`s), `Function` was just for convenience. But I'll update the type for `query`.

w3b6x9: @soedirgo oops, spoke too soon!

inian:
This looks good @soedirgo. Looks like we are pushing to a new repo in dockerhub. So as long as we don't delete the old docker image during the transition, I can migrate the middleware and docker-compose-local.yml bit by bit.

kiwicopple (approved):
🔥

PR #92, opened 2021-02-04
fix: broken documentation URL

0xflotus:

docs update

Broken link to documentation


w3b6x9:
@0xflotus good catch, thanks!

PR #93, opened 2021-02-05
fix: remove oid-to-name conversions

soedirgo:

Bug fix.

Turns out that `oid :: reg* :: text` leaves behind some quotes: https://stackoverflow.com/questions/55981426/how-to-convert-regclass-to-string-without-quotes.

We don't want this, especially when doing filtering or showing it to the frontend, so we should use `name` columns from the corresponding system catalogs instead.


kiwicopple (approved):
lgtm @soedirgo 👍

PR #98, opened 2021-02-27
fix(publications): disambiguate tables' schemas

### LINKED ISSUE #97: Selecting Publications with multiple schemas returns duplicate table names
# Bug report

## Describe the bug

image

## To Reproduce

- Create a publication
- create a table "public.users"
- create a table "auth.user"
- Add both tables to the publication
- select the publication

The result will return an array of tables `["users", "users"]`

## Expected behavior

Some way to distinguish which schema the table belong to. Perhaps:

```
[
  { id: X, name: "users", schema: "public" },
  { id: Y, name: "users", schema: "auth" },
]
```

soedirgo:

Bug fix.

`/publications`'s tables only have names, so we can't differentiate tables from different schemas.

## What is the new behavior?

Make `tables` a JSON object as per #97. (Breaking change)

## Additional context

Closes #97.

### DIFF
--- a/src/lib/sql/publications.sql
+++ b/src/lib/sql/publications.sql
@@ -15,15 +15,22 @@ FROM
   LEFT JOIN LATERAL (
     SELECT
       COALESCE(
-        array_agg(c.relname :: text) filter (
-          WHERE
-            c.relname IS NOT NULL
+        array_agg(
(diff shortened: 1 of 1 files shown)


kiwicopple:
🔥

PR #99, opened 2021-02-28
Kaizen

soedirgo:

Bug fix + some cleanup.

## What is the new behavior?

- Remove `docs`: these are pretty outdated, we can add back OAS stuff later when we need it.
- Remove unused stuff from `src/server`: these are stubs for using the new lib, but are currently unused.
- Move old stuff to `src/old`: makes it clear which parts are the new lib stuff and which are the old routes.
- Pluralize `PostgresMeta` objects (`PostgresMetaColumns`, etc.)
- Fix `/publications`: forgot to update the logic & tests in #98. 😅 Probably better to specify `tables` using OIDs when creating publications, but this'll do for now.


kiwicopple:
image


will have to trust you on this one 😅

PR #101, opened 2021-03-08
Migrate routes to use the library

soedirgo:

Bug fix + feature.

### Tests: Remove replica_identity from `POST /tables`
Atm the library tries not to do an `update` (`ALTER`) in a `create` (`POST`), so we remove `replica_identity` test from `POST /tables`. In the future we aim for (almost) complete parity of options between `create` and `update`.
#### Always use `Pool` from `node-postgres`
Originally the library allows using the `Client` class from `node-postgres` (by setting `pooled: false`) which is allegedly for when we only need one DB connection that we manage manually, however it turns out that using it directly is not the "happy path", and it's simpler to just use a `Pool` with max size 1 (`pool.end()` ends connections gracefully, among other things--`client.end()` ends ongoing queries abruptly).
#### Support `pool.end()` from the library
Since we're dynamically creating pools in the server, we want the library to be able to end unneeded pools. Note that in the typical use case of one DB connection this isn't necessary--pools are closed when the server stops (CMIIW).
### Server: Remove `src/old` and migrate to lib
Remove the old routes in `src/old` and migrate to use the library. Pool connections to multiple databases with [lru-cache](https://github.com/isaacs/node-lru-cache) (holds 100 most recently used connections).


kiwicopple:
A lot going on here 😅 

Can I suggest in the future making one conceptual change per PR? The smaller the cognitive load for reviewers, the faster it gets merged :)

>  lru-cache 

I haven't dug in, but will the pools stay open for a set period of time? If so, what will happen if there are over 100 concurrent requests during that time? Did you also get a chance to check how much memory this is going to use per-connection?

> Remove replica_identity from POST /tables

Does this mean there is no way to update a `replica_identity` for a table?

soedirgo:
> Can I suggest in the future making one conceptual change per PR?

Yup my bad, could've split the pooling out here... 😅

> will the pools stay open for a set period of time?

The pools will stay open until they get "pushed out" of the cache, i.e. when they're no longer among the 100 most recently used ones, at which point the server will call `pgMeta.end()` to close the pool (the cache is size/length-based rather than time-based, though you can do that too).

I haven't tested this in a realistic environment (I'm not sure how to test it without actually pooling 100 distinct DB's connections), but locally, caching 100 pools, all connected to one local DB, it takes about 13MB memory on idle, so not half bad.

When there are more than 100 concurrent requests (on different DBs), inevitably some pools will get pushed out of the cache, so for those it should behave the way it worked before (no cache). FWIW, I realized that `pg-pool` by default uses 10 max connections, so it's possible that just by setting the max. pool size to 1 would substantially reduce the latency.

The closest thing I found on memory usage per DB connection is [this comment about pgbouncer](https://stackoverflow.com/questions/35387314/multiple-databases-common-connection-pool-using-pgbouncer#comment58612461_35387314): about 4MB, so I thought 400MB (100 connections) should be managable.

soedirgo:
> Does this mean there is no way to update a replica_identity for a table?

You can still do it on `update`/`PATCH`. I only removed it for consistency, as it was the only place `update` is called from `create` (and there was no usage from infra/CLI).

kiwicopple:
> You can still do it on update/PATCH. 

Ah great! 

Do you think we can split out the pooling into another PR? I think it might need some testing on the fringe cases (for example, a transaction is running when it gets pushed out).

soedirgo:
OK, reverted the caching part 👍

src/lib/db.ts
```diff
@@ -1,19 +1,44 @@
...
+  query: (sql: string) => Promise<PostgresMetaResult<any>>
+  end: () => Promise<void>
+} = (config) => {
+  // XXX: Race condition could happen here: one async task may be doing
+  // `pool.end()` which invalidates the pool and subsequently all existing
+  // handles to `query`. Normally you might only deal with one DB so you don't
+  // need to call `pool.end()`, but since the server needs this, we make a
+  // compromise: if we run `query` after `pool.end()` is called (i.e. pool is
+  // `null`), we temporarily create a pool and close is right after.
+  let pool: Pool | null = new Pool(config)
```
kiwicopple, line 18: Interesting. We haven't seen this at all yet as far as I can tell. Perhaps there is somethign smart going on where there is a pool created per-config? 

So that means the race condition would only really happen if 2 different users are using the same DB at the same time?

kiwicopple:
OK nice one @soedirgo ! Feel free to merge whenever :)

src/lib/db.ts
soedirgo: This doesn't really apply without the cache (since we're creating a new pool on each request). But with the cache it's possible that one async function gets a handle to the pool, yield execution through await, then another async function closes the pool through `pool.end()`.

Should be a rare case though as `pool.end()` is only called when the cache disposes the pool, i.e. the pool hasn't been used for a while right when this is happening.

PR #102, opened 2021-03-16
Use Fastify in place of Express

soedirgo:

Feature.

## What is the new behavior?

Use Fastify for the server in place of Express.

There's no logic change, but it's still a big change in terms of LoC, no way around it since code organization is different under Fastify (can't just `s/express/fastify`).

Some next steps (post-March):
- Request & response validation (currently types from request, minus `body`, are checked at compile time--during runtime all bets are off). Will need to export types from the lib.
- Split up tests, it's getting hairy... Maybe use jest?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,6 +45,22 @@
         "regenerator-runtime": "^0.13.4"
       }
     },
+    "@fastify/forwarded": {
+      "version": "1.0.0",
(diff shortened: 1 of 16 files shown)


kiwicopple:
Wow that was fast. Very nice @soedirgo - looks great, very nice code cleanup

Feel free to merge whenever you want 👍

Issue #100, opened 2021-03-01
Feature: Add constraints


kiwicopple:
# Feature request

## Is your feature request related to a problem? Please describe.

It would be nice to be able to specify constraints within bot a table and a column

For example:

```
create table table_name (
  username text,
  
  unique(username)
)

## Describe the solution you'd like

It may need a similar treatment as the default values:

https://github.com/supabase/postgres-meta/blob/f35c38f00866a50b4629ede876454d810f76ccdb/test/integration/index.spec.js#L352



## Additional context

- https://github.com/supabase/supabase/issues/770

wasabigeek:
Hi! I was thinking of taking a stab at this for columns, but wasn’t exactly sure what interface would be desirable. I see there's already a few constraint options (like `is_unique`) in https://github.com/supabase/postgres-meta/blob/develop/src/lib/PostgresMetaColumns.ts, so would this be about allowing a user to specify a generic definition as free text when creating a column?

kiwicopple:
Yeah, in this case we want to start with something as generic as possible, then we can add more specific functionality for anything which isn't covered 👍 

would love any contributions!

juanzgc:
Unique constraints seem to have been added in the view when creating a new table but not in the update a column view or create a new column view. Would be very helpful to add this feature in those two views as well. 

Note: I've included a few bugs around this as well in the following issue. https://github.com/supabase/supabase/issues/592#issuecomment-917852780

DocLabyrinth:
Could it be possible to allow for uniqueness across multiple columns e.g. if you have a table for storing a user's contact details/socials where the columns look something like id (int), user_id (int, foreign key), type (varchar), value (varchar). In that case you'd want to have unique(user_id, type) as a constraint so it wouldn't be possible to add duplicate values for the same type of contact info.

In the UI there's already similar functionality for marking fields as part of the primary key using check boxes. Could it be possible to provide a similar UI for selecting values to be part of a unique constraint? The only downside to that approach would be that it'd only work for one unique constraint per table when I think postgres allows more than one.

soedirgo:
UI stuff aside: yes, we can make constraints be defined on tables the same way e.g. policies and primary keys are. The idea for this feature is for it to be the solution for all kinds of table constraints, including primary keys and `UNIQUE`s.

jojomatik:
[...]

PR #104, opened 2021-04-30
feat(lib): add naive constraint definition

wasabigeek:

Allows users to define a `constraint` when creating a column https://github.com/supabase/postgres-meta/issues/100. Technically a user could specify multiple constraints in the string, not sure if we'd want something more structured? e.g. an array.

### DIFF
--- a/src/lib/PostgresMetaColumns.ts
+++ b/src/lib/PostgresMetaColumns.ts
@@ -89,7 +89,7 @@ export default class PostgresMetaColumns {
     is_primary_key = false,
     is_unique = false,
     comment,
-    constraint,
+    check,
   }: {
     table_id: number
     name: string
@@ -102,7 +102,7 @@ export default class PostgresMetaColumns {
     is_primary_key?: boolean
     is_unique?: boolean
     comment?: string
-    constraint?: string
+    check?: string
   }): Promise<PostgresMetaResult<PostgresColumn>> {
     const { data, error } = await this.metaTables.retrieve({ id: table_id })
     if (error) {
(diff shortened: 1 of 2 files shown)


wasabigeek:
Would appreciate feedback @kiwicopple @soedirgo 🙏

kiwicopple:
This look awesome @wasabigeek 

> Technically a user could specify multiple constraints in the string, not sure if we'd want something more structured? e.g. an array.

@soedirgo thoughts? From our UI perspective it might be easier

soedirgo:
Thanks for the PR @wasabigeek!

The `is_unique` field in `create()` was there just for convenience since it's part of the `ADD COLUMN` statement, but I think in the future we should be handling constraints in `/constraints`. I thought about just making it a property of a column but constraints aren't only for columns or even tables. I'm not sure what a good API for `/constraints` would look like yet.

So for now we can just support `CHECK` specifically (I think it's better DX to usually have one way of doing things), in which case my suggestions are:
- changing `constraint?: string` to `check?: string`
- changing `constraintSql` to `checkClause`: `` check === undefined ? '' : `CHECK ${check}` ``

> Technically a user could specify multiple constraints in the string, not sure if we'd want something more structured? e.g. an array.

If we're just doing `CHECK` one expression is fine, since it evaluates to a boolean multiple checks can be collapsed to one check.

Thoughts?

wasabigeek:
> it's better DX to ... have one way of doing things

👏 

I like the idea of constraints being it's own "resource". Maybe we could do something like:
```
GET /constraints/?type=check

[
  {
    type: “check”, // not_null, unique, primary_key, foreign_key, exclusion
    definition: “(price > 0)”,
    target_type: “table”, // “column”
    target_id: “...”
  }
]
```
(Not sure offhand if `type` is a reserved keyword)

If we do the above, would we still want to have the `check` attribute on a Column? Maybe it could then be an array of constraint objects instead, or we make them retrieved as part of a separate request? (As an aside, it's possible to assign multiple CHECK constraints to a single column)

soedirgo:
Btw, my reference for what goes into `/constraints` is the [pg_constraint](https://www.postgresql.org/docs/current/catalog-pg-constraint.html) system catalog. So for starters we can have fields that take care of checks, foreign keys, primary keys, and unique constraints. We can discuss this in another issue/PR.

My idea was for this PR to add `CHECK` on column creation, which is a small and easy win. Right now we don't even return the constraints when you retrieve the column.

wasabigeek:
Change as requested @soedirgo!

soedirgo:
Looks good, thanks again for the PR! 😄

PR #105, opened 2021-05-05
test: Add test for creating primary_key column

wasabigeek:
## What kind of change does this PR introduce?

Noticed that there weren't tests for `is_primary_key`, added one. If this approach looks good, I'll try to figure out how to add one for the unique constraint as well.

PR #108, opened 2021-05-10
test: Add test for is_unique columns

wasabigeek:

Adds a test for `is_unique`. Follow-up to #105.

### DIFF
--- a/test/integration/index.spec.js
+++ b/test/integration/index.spec.js
@@ -312,9 +312,7 @@ describe('/tables', async () => {
       type: 'int2',
       default_value: 42,
       is_nullable: false,
-      comment: 'foo',
-      // Currently no way to test these:
-      // is_unique: true
+      comment: 'foo'
     })
 
     const { data: columns } = await axios.get(`${URL}/columns`)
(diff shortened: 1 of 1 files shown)


test/integration/index.spec.js
```diff
@@ -356,6 +354,33 @@ describe('/tables', async () => {
...
+      type: 'int2',
+      is_unique: true,
+    })
+
+    const { data: uniqueColumns } = await axios.post(
+      `${URL}/query`,
+      { query: `
+        SELECT a.attname
+        FROM   pg_index i
+        JOIN   pg_constraint c ON c.conindid = i.indexrelid
```
wasabigeek, line 371: Technically speaking looking at the index alone should be enough (PG uses a [unique index](https://postgresql.org/docs/13/interactive/indexes-unique.html) to enforce a unique constraint) but I added this for completeness, since we specifically want to create a constraint.

kiwicopple:
Great job @wasabigeek. Really appreciate your help here. If you want to continue contributing to this repo just let me know! There are a lot of things planned (type/schema generators would be amazing)

wasabigeek:
@kiwicopple I'd be open to help! Just might need some guidance here and there :)

Issue #115, opened 2021-06-14
Support for Functions


kiwicopple:
# Feature request

## Is your feature request related to a problem? Please describe.

We should implement `trigger` support. This is something we will need to build into our Dashboard

## Describe the solution you'd like

A clean wrapper around `select * from pg_proc`

soedirgo:
What does the requirement look like right now? I think supporting `receive` and `delete` are low-hanging fruits, but info about parameters are trickier:
- they can have optional names
- they can have optional default values
- they can be `in`, `out`, `inout`, `variadic`, `table`

Also, should this contain other kinds of "functions", i.e. aggregate functions, window functions, and [procedures](https://www.postgresql.org/docs/current/xproc.html)? DataGrip for example groups all this under "routines"—should we separate them or combine them under `/routines`?

@i-pip would love for you to chime in as well (and thanks again for the PR!).

kiwicopple:
> should this contain other kinds of "functions", i.e. aggregate functions, window functions, and procedures? 

I don't think so for now. It's not important for our use-case (user defined functions).

> info about parameters are trickier

Understood. The "MVP" for this is to allow a user to create a function, then update/delete it afterwards. So ideally it would need to support these parameters.

This will be required in conjunction with triggers since triggers usually call a "function returning a trigger" like [this](https://github.com/supabase/postgres-meta/pull/116/files#diff-2f66a387f2bc57c08cd4fe3afc55ead81141f1ab15b39bdaa240da7d45a2d1d7). 

For additional context, Steve is working on a library which could call a `faas` from Postgres (like Lambda). The ideal use-case here would be to combine with a trigger.

i-pip:
Thanks for the invitation to chime in @soedirgo . In the create function I made no assumptions about how params are structured, they're simply a string array - this means that the details of whether they are `in` `out` `inout` `variadic` or `table` are left up to the user. 

So a user could provide only the data types e.g. ["int", "int"], names e.g ["a int", "b integer"] or even the whether it's an in or out ["a int", "out result int"]

I also made the assumption that procedures should likely be their own route like "/procedures" 

I like the "/routines" idea - it doesn't even have to break current impl since it can aggregate existing routes

PR #118, opened 2021-06-19
feat(/functions): adds POST /functions, DELETE /functions and PATCH /functions

i-pip:
Implements missing methods (POST & DELETE)

Currently list and get methods are implemented on /functions route -  #115 

## What is the new behavior?

Adds POST /functions (implemented by create method) and DELETE /functions (implemented by remove)

- [x] GET /functions
- [x] GET /functions/:id
- [x] POST /functions
- [x] DELETE /functions/:id
- [x] PATCH /functions/:id
  Function properties that can be changed
    - [x] Name
    - [x] Schema


test/integration/index.spec.js
```diff
@@ -157,6 +157,23 @@ describe('/types', () => {
...
+    rettype: 'integer',
+    language: 'sql',
+  }
+  before(async () => {
+    await axios.post(`${URL}/query`, {
+      query: `DROP FUNCTION IF EXISTS "${func.name}";`,
+    })
+    await axios.post(`${URL}/query`, {
+      query: `CREATE SCHEMA IF NOT EXISTS test_schema;`,
+    })
```
w3b6x9, line 175: @i-pip let's drop this schema when this test suite is complete.

w3b6x9 (changes requested):
@i-pip awesome job on the PR so far! Left you a couple of comments. Please let me know if you have any questions or concerns regarding my review and I'll be here to answer. We would love to merge in these changes as soon as possible!

i-pip:
Thank you for all the feedback @w3b6x9 🙌  making the requested changes now - I love it here!

src/lib/PostgresMetaFunctions.ts
```diff
@@ -57,4 +63,101 @@ export default class PostgresMetaFunctions {
...
+      RETURNS NULL ON NULL INPUT;
+      `
+    const { error } = await this.query(sql)
+    if (error) {
+      return { data: null, error }
+    }
+    return await this.retrieve({ name, schema })
+  }
+
+  async update(
```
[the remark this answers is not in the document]
w3b6x9: @i-pip updating function definition with `CREATE OR REPLACE FUNCTION` might get tricky (i.e. function dropped right before `CREATE OR REPLACE FUNCTION` transaction would create an entirely new function) so feel free to skip for now.

src/lib/PostgresMetaFunctions.ts
```diff
@@ -57,4 +63,101 @@ export default class PostgresMetaFunctions {
...
+    let alter = `ALTER FUNCTION ${ident(old!.schema)}.${ident(old!.name)}${
+      params && params.length ? `(${params.join(',')})` : ''
+    }`
+
+    let schemaSql = ''
+    if (schema !== undefined && schema !== old!.schema) {
+      schemaSql = `${alter} SET SCHEMA ${ident(schema)};`
+      alter = `ALTER FUNCTION ${ident(schema)}.${ident(old!.name)}${
+        params && params.length ? `(${params.join(',')})` : ''
+      }`
```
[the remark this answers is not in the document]
i-pip: @w3b6x9  Thanks again for the great feedback. So far I've made all the changes except this one (having a little trouble here). 

Even with the order of `schemaSql` and `nameSql` reversed, I still have to reset `alter` because when I reverse `schemaSql` and `nameSql`, by the time `schemaSql` is ran (after `nameSql`), `ALTER FUNCTION ${ident(old!.name)}` would no longer be valid because `old!.name` would have been replace by the newly provided `name`  

What I've done instead is dropped the if statement and kept your suggestion (swapped the order of `nameSql` & `schemaSql`) but still resetting `alter`

src/lib/PostgresMetaFunctions.ts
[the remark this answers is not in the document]
i-pip: @w3b6x9 I added this to `async create(` instead, this way it's still possible to update the function definition without creating an entirely new function while trying to update with `PATCH`. Problem though is that I think this breaks RESTful conventions because `CREATE OR REPLACE` on `async create(` would be performing an update with `POST`.

test/integration/index.spec.js
```diff
@@ -184,7 +201,38 @@ describe('/functions', () => {
...
+    assert.equal(updated.schema, 'test_schema')
+  })
+  it('DELETE', async () => {
+    await axios.delete(`${URL}/functions/${func.id}`)
+    const { data: functions } = await axios.get(`${URL}/functions`)
+    const stillExists = functions.some((x) => func.id === x.id)
+    assert.equal(stillExists, false, 'Function is deleted')
+    await axios.post(`${URL}/query`, {
+      query: `DROP SCHEMA test_schema;`,
+    })
```
w3b6x9, line 232: @i-pip i recommend moving this to an `after` hook. See: https://github.com/supabase/postgres-meta/blob/f8757bfa295591f010f6f883ba72b8e7a0b36647/test/integration/index.spec.js#L861

test/integration/index.spec.js
```diff
@@ -184,7 +201,38 @@ describe('/functions', () => {
...
+    assert.equal(newFunc.name, 'test_func')
+    assert.equal(newFunc.schema, 'public')
+    assert.equal(newFunc.language, 'sql')
+    assert.equal(newFunc.return_type, 'int4')
+    func.id = newFunc.id
+  })
+  it('PATCH', async () => {
+    const updates = {
+      name: 'test_func_renamed',
+      params: ['integer', 'integer'],
```
w3b6x9, line 215: @i-pip remember to remove:

```suggestion

```

test/integration/index.spec.js
```diff
@@ -184,7 +201,38 @@ describe('/functions', () => {
...
+      schema: 'test_schema',
+      // extension: 'mathlib', // TODO: test patching function extension
```
w3b6x9, line 217: @i-pip and this as well to make it clean:

```suggestion

```

test/integration/index.spec.js
```diff
@@ -184,7 +201,38 @@ describe('/functions', () => {
 
     assert.deepStrictEqual(functionById, functionFiltered)
   })
+  it('POST', async () => {
+    const { data: newFunc } = await axios.post(`${URL}/functions`, func)
+    assert.equal(newFunc.name, 'test_func')
```
w3b6x9, line 206: @i-pip `assert.equal` is deprecated so I recommend updating:

```suggestion
    assert.strictEqual(newFunc.name, 'test_func')
```

w3b6x9 (changes requested):
@i-pip thank you so much for your quick changes to my previous review. I left you some more comments with change requests. Again, please don't hesitate to reach out if anything comes up and thanks for your patience.

Also, [checklist](https://github.com/supabase/postgres-meta/blob/develop/README.md#progress) was just added to the README. Since you did all the hard work, feel free to rebase `develop` and check off functions!

src/lib/PostgresMetaFunctions.ts
```diff
@@ -57,4 +67,86 @@ export default class PostgresMetaFunctions {
...
+      return { data: null, error }
+    }
+    return await this.retrieve({ name, schema })
+  }
+
+  async update(
+    id: number,
+    {
+      name,
+      schema = 'public',
```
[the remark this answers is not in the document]
i-pip: close call! what was I thinking 😅

src/lib/PostgresMetaFunctions.ts
```diff
@@ -57,4 +67,86 @@ export default class PostgresMetaFunctions {
...
+
+    let alter = `ALTER FUNCTION ${ident(old!.name)}`
+    const nameSql =
+      name === undefined || name == old!.name ? '' : `${alter} RENAME TO ${ident(name)};`
+
+    alter = `ALTER FUNCTION ${ident(name)}`
+    const schemaSql =
+      schema === undefined || schema == old!.schema ? '' : `${alter} SET SCHEMA ${ident(schema)};`
+
+    const sql = `BEGIN;${nameSql} ${schemaSql} COMMIT;`
```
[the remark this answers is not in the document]
i-pip: Done! I like it. I initially wanted to do something like this but didn't want to deviate too much from the existing codebase.

test/integration/index.spec.js
i-pip: oops! 😅

test/integration/index.spec.js
i-pip: I just substituted all occurrences of `assert.equal` with `assert.strictEqual` - I hope that's okay

test/integration/index.spec.js
```diff
@@ -14,15 +14,15 @@ console.log('Running tests on ', URL)
 describe('/', () => {
   it('GET', async () => {
     const res = await axios.get(`${URL}/`)
-    assert.equal(res.status, STATUS.SUCCESS)
-    assert.equal(!!res.data.version, true)
+    assert.strictEqual(res.status, STATUS.SUCCESS)
```
w3b6x9, line 17: @i-pip thanks for updating all of these to `strictEqual`!

w3b6x9 (approved):
@i-pip awesome job on this PR! thanks again for the quick turnaround on changes!

PR #123, opened 2021-06-29
Fix: remove double quotes from publications create and update

w3b6x9:

Bug fix

Related issue: https://github.com/supabase/supabase/issues/2123

Getting error: `"relation \"public.test\" does not exist"` (example error) b/c of the double quotes.

## What is the new behavior?

Double quotes are removed and I'm able to create/update publications


soedirgo:
OK, that should do it... thanks again for the PR Wenbo!

PR #127, opened 2021-07-16
fix(functions): include argument_types retrieving function query

w3b6x9:

Bug fix

retrieve function returns multiple records due to overloaded function names.

## What is the new behavior?

retrieve function returns single record despite overloaded function names.

## Additional context

@i-pip creating this PR as it was out of scope for https://github.com/supabase/postgres-meta/pull/118.


soedirgo:
👍 LGTM!

PR #131, opened 2021-07-20
feat: add additional function create options

w3b6x9:

feature

## What is the new behavior?

Add the following options when creating functions:

- [X] IMMUTABLE / STABLE / VOLATILE
- [X] SECURITY INVOKER / SECURITY DEFINER
- [X] SET configuration_parameter { TO value | = value | FROM CURRENT }

### DIFF
--- a/src/lib/PostgresMetaFunctions.ts
+++ b/src/lib/PostgresMetaFunctions.ts
@@ -82,6 +82,9 @@ export default class PostgresMetaFunctions {
     definition,
     rettype = 'void',
     language = 'sql',
+    behavior = 'VOLATILE',
+    security_definer = false,
+    config_params,
   }: {
     name: string
     schema?: string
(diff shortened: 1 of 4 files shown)


soedirgo (approved):
LGTM!

soedirgo:
Actually, one sec - wouldn't you want to allow setting multiple configs for `config_param` and `config_values`?

src/lib/PostgresMetaFunctions.ts
```diff
@@ -82,21 +82,32 @@ export default class PostgresMetaFunctions {
     definition,
     rettype = 'void',
     language = 'sql',
+    behavior,
```
soedirgo, line 85: Can make `VOLATILE` the default here instead

src/lib/PostgresMetaFunctions.ts
```diff
@@ -82,21 +82,32 @@ export default class PostgresMetaFunctions {
...
+    config_values,
   }: {
     name: string
     schema?: string
     args?: string[]
     definition: string
     rettype?: string
     language?: string
+    behavior?: 'IMMUTABLE' | 'STABLE' | 'VOLATILE'
+    security_definer?: 'false' | 'true'
```
soedirgo, line 97: Should we make this a `boolean` instead?

src/lib/sql/functions.sql
```diff
@@ -8,9 +8,34 @@ SELECT
...
   LEFT JOIN pg_language l ON p.prolang = l.oid
   LEFT JOIN pg_type t ON t.oid = p.prorettype
+  LEFT JOIN (
+    SELECT
+	    oid as id,
+	    (string_to_array(unnest(proconfig), '='))[1] AS param,
+   	  string_to_array((string_to_array(unnest(proconfig), '='))[2], ', ') AS values
+	  FROM
+	    pg_proc
+	) p_config ON p_config.id = p.oid
```
w3b6x9, line 32: @soedirgo not sure why the format here is so weird. Looks fine on my local:

Screen Shot 2021-07-20 at 3 14 52 AM

soedirgo: Ah, looks like some tabs starting from line 27. Not a big deal.

PR #134, opened 2021-07-22
feat: add table_id to trigger return

w3b6x9:

feature

## What is the new behavior?

add tg_rel_id (trigger table id) in trigger return


soedirgo:
Can we rename it to `table_id`?

soedirgo:
Looks good 👍

PR #136, opened 2021-07-22
fix: create functions with alias data types

w3b6x9:

Bug fix

create function, but fails to fetch created function, with function args being alias data types like `int2, int4, etc.`

## What is the new behavior?

create function, then successfully fetch created function, with function args being alias data types like `int2, int4, etc.`

## Additional context

test example:

```
{
   "name":"add",
   "schema":"public",
   "definition":"SELECT a + b;",
   "rettype":"integer",
   "language":"sql",
   "behavior":"STABLE",
   "args":["a int4","b int4"]
}
```

### DIFF
--- a/src/lib/PostgresMetaFunctions.ts
+++ b/src/lib/PostgresMetaFunctions.ts
@@ -53,10 +53,31 @@ export default class PostgresMetaFunctions {
         return { data: data[0], error }
       }
     } else if (name && schema && args) {
-      const argTypes = args.join(', ')
-      const sql = `${enrichedFunctionsSql} WHERE schema = ${literal(schema)} AND name = ${literal(
-        name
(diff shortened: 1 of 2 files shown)


src/lib/PostgresMetaFunctions.ts
```diff
@@ -53,18 +53,46 @@ export default class PostgresMetaFunctions {
...
+                split_args.arr,
+                1
+              ) - 1
+            ],
+            split_args.arr[
+              array_length(
+                split_args.arr,
+                1
+              )
+            ]::regtype
```
soedirgo, line 75: I think the indexes can be `1` and `2`? Otherwise looks good 👍

soedirgo (approved):

PR #137, opened 2021-07-25
feat: enable function definition update

w3b6x9:

feature

## What is the new behavior?

Enable function definition update.

### DIFF
--- a/src/lib/PostgresMetaFunctions.ts
+++ b/src/lib/PostgresMetaFunctions.ts
@@ -3,6 +3,18 @@ import { DEFAULT_SYSTEM_SCHEMAS } from './constants'
 import { functionsSql } from './sql'
 import { PostgresMetaResult, PostgresFunction } from './types'
 
+type FunctionInputs = {
+  name: string
+  definition: string
+  args?: string[]
+  behavior?: 'IMMUTABLE' | 'STABLE' | 'VOLATILE'
+  config_params?: { [key: string]: string }
+  schema?: string
+  language?: string
+  return_type?: string
+  security_definer?: boolean
+}
+
 export default class PostgresMetaFunctions {
   query: (sql: string) => Promise<PostgresMetaResult<any>>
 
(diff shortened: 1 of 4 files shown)


src/lib/PostgresMetaFunctions.ts
```diff
@@ -196,6 +212,84 @@ export default class PostgresMetaFunctions {
...
+        config_params
+          ? Object.entries(config_params)
+              .map(
+                ([param, value]) =>
+                  `SET ${param} ${value[0] === 'FROM CURRENT' ? 'FROM CURRENT' : 'TO ' + value}`
+              )
+              .join('\n')
+          : ''
+      }
+      RETURNS NULL ON NULL INPUT
```
soedirgo, line 250: The default is `CALLED ON NULL INPUT` right? Any reason for this one?

soedirgo (commented):
I have other comments re: the implementation but I'll sign off on this, we can always refactor later

src/lib/PostgresMetaFunctions.ts
w3b6x9: @soedirgo i'll make the update to the default

w3b6x9:
@soedirgo 

> I have other comments re: the implementation

How would you have implemented this? I'll go and update this PR if there isn't a huge deviation between our approaches.

soedirgo:
Hmm, for `FunctionInputs`, it'd be better to do it like schemas in lib/types:
https://github.com/supabase/postgres-meta/blob/bcbd59ccfbfc14d979730fecfd2c30f9e0be1e01/src/lib/types.ts#L198-L202
then the arg type for `generateCreateFunctionSql` can be `PostgresFunctionCreate` (`argument_types` -> `args` can be handled outside the function).

Nit: for `terminateCommand`, I think empty statements are harmless so you can just insert the semicolon if unsure.

Re: using `plpgsql`, notice how the codebase in general has a [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) problem, e.g. doing an `update` based on the `id` from `retrieve` (which may be deleted right after the retrieve). I think in the future we should be using procedural plpgsql, so good to see it can work here at least. Also, TIL that `CREATE OR REPLACE` maintains the `id` - nice one! 😄

w3b6x9:
@soedirgo thanks for the feedback! I'm going to merge and deploy these changes so we can continue local development. I'll quickly follow-up with another PR regarding `FunctionInputs` -> `PostgresFunctionCreate` and removing `terminateCommand`.

PR #149, opened 2021-09-15
fix(server): Cannot read property 'accept-version' of undefined

### LINKED ISSUE #144: Cannot read property `accept-version` when changing column from array to non-array
- [x] Repro

soedirgo:

Bug fix.

Some failing requests are returning cryptic `Cannot read property 'accept-version' of undefined` error.

## What is the new behavior?

Sayonara cryptic errors. Responses should now provide the *actual* errors.

## Additional context

Closes #144.
Related: https://github.com/supabase/supabase/issues/2404#issuecomment-913153702

### DIFF
--- a/src/server/routes/tables.ts
+++ b/src/server/routes/tables.ts
@@ -37,7 +37,7 @@ export default async (fastify: FastifyInstance) => {
     const { data, error } = await pgMeta.tables.retrieve({ id })
     await pgMeta.end()
     if (error) {
-      request.log.error({ error, req: request.body })
+      request.log.error(JSON.stringify({ error, req: request.body }))
       reply.code(404)
       return { error: error.message }
     }
--- a/package.json
+++ b/package.json
@@ -21,11 +21,11 @@
     "build:server": "tsc -p tsconfig.server.json && cpy 'src/lib/sql/*.sql' bin/src/lib/sql",
     "docs:export": "PG_META_EXPORT_DOCS=true ts-node-dev src/server/app.ts",
     "start": "NODE_ENV=production node bin/src/server/app.js",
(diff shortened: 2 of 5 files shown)


src/server/routes/tables.ts
```diff
@@ -37,7 +37,7 @@ export default async (fastify: FastifyInstance) => {
     const { data, error } = await pgMeta.tables.retrieve({ id })
     await pgMeta.end()
     if (error) {
-      request.log.error({ error, req: request.body })
+      request.log.error(JSON.stringify({ error, req: request.body }))
```
darora, line 40: This...fixes the response sent back to the client?

soedirgo: Ostensibly yeah, `request.log.error` accepts a string, not sure what the handler does but once I pass it an object all bets are off

soedirgo: Not sure where the `accept-version` stuff comes from.

soedirgo: Interesting, in fact setting `req` to anything triggers the error. Doesn't happen when I set e.g. `a`. Not sure why that matters.

inian: Ahh makes sense now. I think fastify sets that key and that key is overriden by you here. So the internal logging module is trying to do req.headers.accept-version and that ends up throwing the error.

soedirgo: `JSON.stringify`ing `request.body` seems fine. Either way simply passing a string as per docs seems to have fixed it, so probably not worth it to go further down the rabbit hole.

darora (approved):

PR #159, opened 2021-11-21
feat: add a parser and formatter

kiwicopple:

Adds a parser and formatter

## Additional context

We're using the `node-sql-parser` lib in our frontend app and it's a huge PITA to build. Shifting it here makes a lot of sense (it's on a single API route in `studio`)


src/lib/Parser.ts
```diff
@@ -0,0 +1,59 @@
...
+  database: 'PostgreSQL',
+  type: 'table',
+}
+
+const DEFAULT_FORMATTER_OPTIONS = {
+  plugins: [SqlFormatter],
+  formatter: 'sql-formatter',
+  language: 'postgresql',
+  database: 'postgresql',
+  uppercase: false, // I thought this would "lowercase" everything, but it doesn't. There is not option to convert everything to lowercase.
```
kiwicopple, line 16: I would love an option to lowercase the values. it would require a fork of this: https://github.com/zeroturnaround/sql-formatter (which hasn't been active since Feb 2021)

src/lib/Parser.ts
```diff
@@ -0,0 +1,59 @@
...
+  parser: 'sql',
+}
+
+/**
+ * Parses a SQL string into an AST.
+ */
+export function Parse(sql: string, options: ParserOptions = {}): ParseReturnValues {
```
kiwicopple, line 23: Ideally the parser would also include the ability to replace patterns so that we can do templates (similar to Redash)

soedirgo (approved):

PR #165, opened 2021-12-15
fix: improve logging

soedirgo:
Had to dig through pg-meta logs through CloudWatch and it was next to useless. This improves the log output by:
- reverting to log error, request, etc. as JSON instead of stringified JSON (from https://github.com/supabase/postgres-meta/pull/149)
- logs the db hostname the query is run against

@darora let me know if this is violating personal info thingy or if there's any extra stuff you want logged!

### DIFF
--- a/src/server/utils.ts
+++ b/src/server/utils.ts
@@ -5,5 +5,4 @@ export const extractRequestForLogging = (request: FastifyRequest) => ({
   method: request.method,
   url: request.url,
   pg: parse(request.headers.pg as string).host,
-  body: request.body,
 })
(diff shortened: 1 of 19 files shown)


src/server/utils.ts
```diff
@@ -0,0 +1,9 @@
+import { parse } from 'pg-connection-string'
+import { FastifyRequest } from 'fastify'
+
+export const extractRequestForLogging = (request: FastifyRequest) => ({
+  method: request.method,
+  url: request.url,
+  pg: parse(request.headers.pg as string).host,
+  body: request.body,
```
darora, line 8: Hmm I'm not sure if it's OK to log the body by default. Aside from concerns about the size of the object that might get logged, we might also end up logging sensitive data here.

If needed, we could log a hash of the body?

darora (approved):

src/server/utils.ts
soedirgo: :+1: removed `body` for now

PR #176, opened 2022-01-26
fix: #143, #174, #175

### LINKED ISSUE #143: Support results for multi-statement queries

### LINKED ISSUE #174: `DATE`s are returned as JS `Date`s
Context: https://github.com/supabase/supabase/issues/4076

### LINKED ISSUE #175: Columns with the same alias get merged
https://node-postgres.com/api/result

soedirgo:
## Fixes #174

`date`s, `timestamp`s, and `timestamptz`s are no longer converted to JS `Date`s.

`date` before: `"2020-01-01T00:00:00.000Z"`

`date` after: `"2020-01-01"`

## Fixes #143

Queries w/ multiple statements now return results from all statements instead of returning an empty list.

`select 1 as a; select 2 as b` before:
```json
[]
```
`select 1 as a; select 2 as b` after:
```json
[
  {
    "columns": [
      "a"
    ],
    "rows": [
      [
        1
      ]
    ]
  },
  {
    "columns": [
      "b"
    ],
    "rows": [
      [
        2
      ]
    ]
  }
]
```

## Fixes #175

Columns with the same alias now works, but this required a breaking change in the response format.

`select 1 as x, 2 as x` before:
```json
[
  {
    "x": 2
  }
]
```

`select 1 as x, 2 as x` after:
```json
[
  {
    "columns": [
      "x",
      "x"
    ],
    "rows": [
      [
        1,
        2
      ]
    ]
  }
]
```

## Additional notes

Previously, the response is a list of rows, but now that we support multi-statement queries, the response is a list of query results. In other words:

Type of response before: `Record<string, unknown>[]`
Type of response after: `{ columns: string[]; rows: unknown[][] }[]`


soedirgo:
In hindsight, I wonder if this would require a lot of refactors in Studio. If there are a lot of call sites that rely on the old response format, we can expose the "arrayMode" as a new endpoint instead.

PR #181, opened 2022-02-08
fix: handling of type identifier in column create() and update()

DukeManh:

Fix [issue 5330](https://github.com/supabase/supabase/issues/5330) of Supabase Upstream. 

Updating column data type from `type` to `type[]`  fails because of insufficient identifier escaping in colum `create` and `update` function.

## Additional context

This is actually a TO-DO  so it's safe to delete the Todo comment now :).


DukeManh:
Can you please review, @kiwicopple, @soedirgo , @w3b6x9

soedirgo:
This is great, thanks @DukeManh!

soedirgo (approved):

PR #183, opened 2022-02-18
fix: date & timestamp format

soedirgo:
Fixes https://github.com/supabase/postgres-meta/issues/174.

dates, timestamps, and timestamptzs are no longer converted to JS Dates.

date before: "2020-01-01T00:00:00.000Z"

date after: "2020-01-01"


soedirgo:
I think this one's simple & doesn't require a change on the dashboard

PR #184, opened 2022-02-24
Reduce docker image size

ffabss:

Reduces the total size of the docker image by using node:14-alpine as the base image.

The image is based on node:14.

## What is the new behavior?

The docker image is smaller (now ~341 MB).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,23 @@
-FROM node:14
+FROM node:14-alpine as deps
+RUN apk add --no-cache bash make git python3
+RUN apk add --update alpine-sdk
 WORKDIR /usr/src/app
 # Do `npm ci` separately so we can cache `node_modules`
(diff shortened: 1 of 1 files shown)


soedirgo:
This reduces the final image size if you're only using this image in your system, but actually increases it in the common case because your images wouldn't share the same layers. See also https://github.com/nodejs/docker-node/blob/main/README.md#nodeversion.

400+ MB is not great though. A better approach for reducing image size would be doing multi-stage builds, as in https://github.com/supabase/stripe-sync-engine/blob/main/Dockerfile. Can you check how well that works?

soedirgo:
My bad, somehow I missed the fact that it's already multi-stage...

Can you use `node:14` just for the build stage? And no need for separating `deps` and `build` stages.

soedirgo:
Actually, scratch that, this might be even better. Thanks for the PR!

soedirgo (approved):

PR #206, opened 2022-03-19
[wip] feat: batch endpoints for column creation and retrieval

darora:
existing tests pass


src/server/routes/columns.ts
```diff
@@ -54,6 +55,26 @@ export default async (fastify: FastifyInstance) => {
     return data
   })
 
+  fastify.get<{
```
darora, line 58: I've opted to expose these separately rather than have the existing endpoints handle both cases; it gets annoying with the GET accepting a single ID param (and just gnarly in any case).

We can replace the current endpoint as part of a breaking change in the future.

soedirgo (approved):

src/lib/PostgresMetaColumns.ts
```diff
@@ -57,75 +57,148 @@ export default class PostgresMetaColumns {
...
+    names?: string[]
+    table?: string
+    schema?: string
+  }): Promise<PostgresMetaResult<PostgresColumn[]>> {
+    if (ids && ids.length > 0) {
       const regexp = /^(\d+)\.(\d+)$/
-      if (!regexp.test(id)) {
-        return { data: null, error: { message: 'Invalid format for column ID' } }
+
+      const invalidId = ids.find((id) => !regexp.test(id))
```
darora, line 102: Would recommend findAll (or equivalent) here

src/lib/PostgresMetaColumns.ts
```diff
@@ -57,75 +57,148 @@ export default class PostgresMetaColumns {
...
+    table: string
+    schema: string
+  }): Promise<PostgresMetaResult<PostgresColumn[]>>
+  async batchRetrieve({
+    ids,
+    names,
+    table,
+    schema = 'public',
+  }: {
+    ids?: string[]
```
darora, line 94: I really dislike this way of everything being optional - can we use a discriminated union type instead?

src/server/routes/columns.ts
```diff
@@ -56,22 +62,49 @@ export default async (fastify: FastifyInstance) => {
...
+  }>(
+    '/',
+    {
+      schema: {
+        headers: Type.Object({
+          pg: Type.String(),
+        }),
+        body: Type.Union([postgresColumnCreateSchema, Type.Array(postgresColumnCreateSchema)]),
+        response: {
+          200: Type.Union([postgresColumnSchema, Type.Array(postgresColumnSchema)]),
```
darora, line 75: If we wanted to be really nice about it, would be cool to pluck the table name out of the array, since we don't allow multiple different values for the field here anyway

darora (commented):
Looks good! Couple of minor suggestions

src/lib/PostgresMetaColumns.ts
soedirgo: It's a TS thing - the 2 function declaration above it means it's either `ids` or `names` + `table` + `schema`

src/lib/PostgresMetaColumns.ts
soedirgo: :+1:

src/server/routes/columns.ts
soedirgo: I think it's better to make the response format consistent: `T | T[]` for everything

PR #225, opened 2022-04-05
feat: allow end user to turn on ssl mode for db connections

cpdeethree:

Allows end user to optionally enable SSL mode on db connections 

Currently the behavior is to disallow all SSL connections and not provide end user the option to allow them

## What is the new behavior?

New behavior is to default to disallow SSL connections but allow end user to allow them via environment variable

## Additional context

While this app is, as the readme states, intended to run in an unexposed environment, there are still valid use cases for having SSL enabled in these kinds of environments.

### DIFF
--- a/src/server/constants.ts
+++ b/src/server/constants.ts
@@ -6,10 +6,11 @@ const PG_META_DB_NAME = process.env.PG_META_DB_NAME || 'postgres'
 const PG_META_DB_USER = process.env.PG_META_DB_USER || 'postgres'
 const PG_META_DB_PORT = Number(process.env.PG_META_DB_PORT) || 5432
 const PG_META_DB_PASSWORD = process.env.PG_META_DB_PASSWORD || 'postgres'
+const PG_META_DB_SSL_MODE = process.env.PG_META_DB_SSL_MODE|| 'disable'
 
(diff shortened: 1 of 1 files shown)


soedirgo:
Maybe `PG_CONNECTION` should be named `DEFAULT_PG_CONNECTION` instead to avoid confusion, but you're meant to pass an `x-connection-encrypted` header to specify the DB URL per request. You can enable ssl this way.

cpdeethree:
Makes sense - but the main upstream project that uses this (Supabase) just uses and exposes the default connection string for downstream users: https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L147

And I'm using that via the community helm chart: https://github.com/supabase/supabase/blob/master/docker/docker-compose.yml#L147

Should I ask there if it makes more sense to pass authentication via the headers? Seems like it might make configuring things a bit more difficult that way in downstream projects that call this and might be a much bigger change

soedirgo:
Ah, I see - seems like it's easier to add this option instead. If you don't mind resolving the prettier check I'll merge this in :+1:

cpdeethree:
fixed by running prettier 👍

soedirgo (approved):

PR #260, opened 2022-05-30
POC: typescript generator

kiwicopple:
Draft/POC

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "@sinclair/typebox": "^0.23.4",
+    "ejs": "^3.1.8",
     "pg": "^8.7.1",
     "pg-format": "^1.0.4",
     "pgsql-parser": "^13.3.0",
@@ -45,6 +46,7 @@
   },
   "devDependencies": {
     "@types/crypto-js": "^4.1.1",
+    "@types/ejs": "^3.1.1",
     "@types/jest": "^27.0.1",
     "@types/node": "^17.0.21",
     "@types/pg": "^8.6.5",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -62,7 +64,7 @@
     "pkg": "^5.5.2",
     "rimraf": "^3.0.2",
     "ts-jest": "^27.0.5",
-    "ts-node-dev": "^1.1.6",
+    "ts-node-dev": "^2.0.0",
```
kiwicopple, line 67: I'm on node v16 now - needed to bump this

src/lib/templates/typescript.ts
```diff
@@ -0,0 +1,20 @@
...
+      <% functions.forEach(function(fn){ %>
+          <%= fn.name %>: {},
+      <% }); %>
+    }
+  }
+}
+
+`.trim()
+
+export default template
```
kiwicopple, line 20: This is the main template, based on the spec in the RFC

src/server/routes/generators/typescript.ts
```diff
@@ -0,0 +1,22 @@
...
+    Headers: { pg: string }
+  }>('/', async (request, reply) => {
+    const connectionString = request.headers.pg
+    const responseFormat = request.headers.responseFormat || 'text/plain'
+
+    const pgMeta = new PostgresMeta({ ...DEFAULT_POOL_CONFIG, connectionString })
+
+    const { data: tables } = await pgMeta.tables.list({})
+    const { data: functions } = await pgMeta.functions.list({})
+    const definition = ejs.render(template, { tables, functions })
```
kiwicopple, line 18: pgMeta already has the data in JSON format, so we have done the hard work - we just need to pass the data to the template

src/server/routes/generators/typescript.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    return reply.header('content-type', responseFormat).send(definition)
```
kiwicopple, line 20: then we can send the data in text/plain, so that the CLI can dump it to disk (or we can display it on the Dashboard etc)

Result:

image

soedirgo: Do we need configurable `responseFormat`? Or can we make it `text/plain` always?

soedirgo:
Closed in favor of #269

PR #280, opened 2022-06-23
feat: shoehorn CLI-mode

soedirgo:
Used for docs export & typegen.

Probably need to move it to `src/cli` in the future, but works well enough for now.


soedirgo:
I'll merge this for now to unblock the CLI

PR #330, opened 2022-08-23
fix: escape db password

### LINKED ISSUE #323: Escape postgres connection string
https://github.com/supabase/postgres-meta/blob/32da58fb3e18d21bd15ab423bb04a0267275d8ea/src/server/constants.ts#L15

special characters in `PG_META_DB_PASSWORD` need to be escaped.

related to https://github.com/supabase/cli/issues/372

soedirgo:
Fixes #323


sweatybridge (approved):
Do we want to regex parse `PG_META_DB_URL` to escape the password there?

soedirgo:
Nah this is fine, I don't trust my own regex skills 😛

Issue #332, opened 2022-08-28
Type error: Don't use `{}` as a type


probablykasper:
# Bug report

## Describe the bug

With v2 I get this error after having generated types, probably because of TS strict mode:
```
Don't use `{}` as a type. `{}` actually means "any non-nullish value".
- If you want a type meaning "any object", you probably want `Record<string, unknown>` instead.
- If you want a type meaning "any value", you probably want `unknown` instead.
- If you want a type meaning "empty object", you probably want `Record<string, never>` instead.
```

due to the generated types including this:
```ts
export interface Database {
		Functions: {}
	}
}
```

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

TypeScript strict mode probably needs to be enabled.
Run `supabase gen types typescript --local > databaseDefinitions.ts`

## System information

- OS: macOS
- Version of supabase-js: 2.0.0-rc.6
- Version of Node.js: 16.15.0

soedirgo:
The error seems to be coming from typescript-eslint. What do you suggest the replacement should be? `Record<string, never>` is not exactly correct, see https://github.com/microsoft/TypeScript/issues/47486#issuecomment-1015671856

probablykasper:
I don't really see why `Record<string, never>` doesn't fit. It doesn't seem like it's supposed to have any properties, so why would you ever need to access it's properties?

soedirgo:
It needs to give a TS error when accessing a property, whereas `Record<string, never>` allows accessing properties (with value of type `never`).

soedirgo:
Actually, `Record<never, never>` seems to work. It resolves to `{}` but seems to fool ts-eslint somehow.

PR #334, opened 2022-08-30
fix(typegen): change empty object type

soedirgo:
Change `{}` to `{ [_ in never]: never }`.

Fixes https://github.com/supabase/postgres-meta/issues/332.


src/server/app.ts
```diff
@@ -92,7 +92,7 @@ if (EXPORT_DOCS) {
             GENERATE_TYPES_INCLUDED_SCHEMAS.includes(name)
         ),
         tables,
-        functions,
+        functions: functions.filter((f) => !['trigger', 'event_trigger'].includes(f.return_type)),
```
thebengeu, line 95: Optional: could destructure `return_type`, here and below.

thebengeu (approved):

src/server/app.ts
soedirgo: ahh true

PR #336, opened 2022-08-31
feat: support docker file-based secrets

burmecia:

This PR is to add an entry point file `docker-entrypoint.sh` to Docker image, which will enable use of Docker file-based secrets as mentioned in https://github.com/supabase/supabase/issues/6014.

Currently this Docker image only supports environment variable based secrets, file-based secrets `*_FILE` are not supported but is suggested by Docker in https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images.

## What is the new behavior?

File-based Docker secrets `*_FILE` will be supported and fall back to environment variable secrets.

## Additional context

N/A

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+# usage: file_env VAR [DEFAULT]
+#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
+# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
+#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
+file_env() {
+	local var="$1"
+	local fileVar="${var}_FILE"
+	local def="${2:-}"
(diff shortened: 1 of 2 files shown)


docker-entrypoint.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+# usage: file_env VAR [DEFAULT]
+#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
+# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
+#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
```
sweatybridge, line 7: Quoting from the referenced issue:

> Some people use tricks to read these files and export to environment variables, but it's not recommended:
>
> Docker secrets do not set environment variables directly. This was a conscious decision, because environment variables can unintentionally be leaked between containers (for instance, if you use --link).

Shall we read the file from nodejs app instead?

sweatybridge: adding native support, ptal: https://github.com/supabase/postgres-meta/pull/337

PR #337, opened 2022-08-31
feat: load secret from files

sweatybridge:
as discussed in https://github.com/supabase/postgres-meta/pull/336

- [x] depends on https://github.com/supabase/postgres-meta/pull/442


src/server/constants.ts
```diff
@@ -1,11 +1,13 @@
+import { getSecret } from '../lib/secrets'
+
 export const PG_META_PORT = Number(process.env.PG_META_PORT || 1337)
-export const CRYPTO_KEY = process.env.CRYPTO_KEY || 'SAMPLE_KEY'
+export const CRYPTO_KEY = (await getSecret('CRYPTO_KEY')) || 'SAMPLE_KEY'
```
soedirgo, line 4: Top-level await is only available in ESM. For one-off use cases like this using the sync functions should be fine.

src/server/constants.ts
```diff
@@ -1,11 +1,13 @@
...
 
 const PG_META_DB_HOST = process.env.PG_META_DB_HOST || 'localhost'
 const PG_META_DB_NAME = process.env.PG_META_DB_NAME || 'postgres'
 const PG_META_DB_USER = process.env.PG_META_DB_USER || 'postgres'
 const PG_META_DB_PORT = Number(process.env.PG_META_DB_PORT) || 5432
-const PG_META_DB_PASSWORD = process.env.PG_META_DB_PASSWORD || 'postgres'
+const PG_META_DB_PASSWORD = (await getSecret('PG_META_DB_PASSWORD')) || 'postgres'
```
soedirgo, line 10: Ditto

src/server/constants.ts
sweatybridge: What's wrong with using ESM?

soedirgo: Nothing wrong with it, but if we want it we should do the switch in another PR.

soedirgo: > And our package.json also specifies the module type

Are you sure? It's not specified there, and iirc the default is `commonjs`.

soedirgo (approved):

PR #360, opened 2022-09-23
feat: add config support for role

fuergaosi233:

Feat

Added the option to set Config at creation time, which we can now use to manage the Config of the Role, such as search_path

## What is the new behavior?
We can set role when create.
image

### DIFF
--- a/src/lib/PostgresMetaRoles.ts
+++ b/src/lib/PostgresMetaRoles.ts
@@ -2,7 +2,22 @@ import { ident, literal } from 'pg-format'
 import { DEFAULT_ROLES } from './constants'
 import { rolesSql } from './sql'
 import { PostgresMetaResult, PostgresRole } from './types'
-
+export interface PostgresMetaRoleConfig {
+  // https://www.rfc-editor.org/rfc/rfc6902
+  op: 'remove' | 'add' | 'replace'
+  path: string
+  value?: string
+}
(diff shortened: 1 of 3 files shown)


fuergaosi233:
@soedirgo Pls help me review this pr 🙏

fuergaosi233:
Reping @soedirgo

soedirgo:
Hey there, a couple things:

- Can you check what happens when the of a config is quoted? I see that you do a `literal(v)` there - I don't think that's correct, e.g. for `search_path="extention,public"` in your test (that's a typo btw) it's interpreted as a single element named `"extention,public"`, rathen than two elements `extention` and `public`.
- Can we turn `null`s into `[]` so the type is consistent?
- Can we remove `"op": "test"`? That seems unused, afaict we only need `"add"`, `"replace"`, and `"remove"`.

fuergaosi233:
1. I remove literal
2. Done
3. Why we need change `null` 2 `[]`, i don't get point.
4. Done

soedirgo:
Thanks for the PR!

> Why we need change null 2 []

This would be `{}` now instead of `[]` - the reason is to make the type simpler, since `Record<string, string>` can already represent an empty `config`, so we don't need it to be `Record<string, string> | null`.

soedirgo (approved):

PR #375, opened 2022-10-07
feat: filtering by schema(s) for tables, columns, types, policies and functions

### LINKED ISSUE #178: Pass Excluded Schemas When Retrieving List of All Functions
# Feature request

## Is your feature request related to a problem? Please describe.

When a couple of extensions have been enabled, and they come with functions, then I can have hundreds of functions in a database. I want to return only a list of functions in schemas I care about, not all functions in all schemas (like `extensions`, `realtime`, etc.). This will also improve API response performance.

Pass a list of schemas I wish to exclude from the functions list query.

osaxma:

This PR introduces a new feature that allows filtering by schema(s) for the following:
- tables
- columns
- types
- policies
- functions (fixes #178)

Currently, it's not possible to filter by schema(s).

## What is the new behavior?

This PR allows filtering the given endpoints with a list of schemas using either `included_schemas` or `excluded_schemas` as a query parameter.

Example Usage:

```
curl 'http://0.0.0.0:1337/tables?included_schemas=public'
curl 'http://0.0.0.0:1337/columns?included_schemas=public,core,private'
curl 'http://0.0.0.0:1337/functions?excluded_schemas=public'
curl 'http://0.0.0.0:1337/types?excluded_schemas=pg_catalog&include_system_schemas=true'
```

## Additional context

Currently, it's only possible to pass schema(s) as a comma separated values (i.e., `"...?included_schemas=public,core"`). 

When I tried defining `included_schemas` or `excluded_schemas` as `string[]` such as:

```ts
  fastify.get<{
    ....
    Querystring: {
      included_schemas?: string[] 
      excluded_schemas?: string[]
    }
   ....
  }>
``` 
Single value and comma separated values were still parsed as a string. This type argument would parse an array only if the parameter was passed multiple times (e.g. `"...?included_schemas=public&included_schemas=core"`). Based on a quick research, it seems arrays can be parsed correctly when using JSON Schema instead of generics but I didn't want to complicate this PR. So the API with this change will be limited to comma separated values for `included_schemas` and `excluded_schemas` parameters.


sweatybridge (approved):

sweatybridge:
I've updated your PR with support for triggers and minor refactoring. Supersedes https://github.com/supabase/postgres-meta/pull/440

osaxma:
thanks @sweatybridge !

PR #423, opened 2022-11-25
fix(typegen): function with table return type

soedirgo:
addresses https://github.com/supabase/cli/issues/635, https://github.com/supabase/supabase/discussions/10464

### DIFF
--- a/src/lib/PostgresMetaTypes.ts
+++ b/src/lib/PostgresMetaTypes.ts
@@ -11,23 +11,35 @@ export default class PostgresMetaTypes {
   }
 
   async list({
+    includeArrayTypes = false,
     includeSystemSchemas = false,
     limit,
     offset,
   }: {
+    includeArrayTypes?: boolean
     includeSystemSchemas?: boolean
     limit?: number
     offset?: number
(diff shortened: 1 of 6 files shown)


src/lib/PostgresMetaTypes.ts
```diff
@@ -11,23 +11,35 @@ export default class PostgresMetaTypes {
...
   }: {
+    includeArrayTypes?: boolean
     includeSystemSchemas?: boolean
     limit?: number
     offset?: number
   } = {}): Promise<PostgresMetaResult<PostgresType[]>> {
     let sql = typesSql
+    if (!includeArrayTypes) {
+      sql += ` and not exists (
+                 select
```
sweatybridge, line 27:
```suggestion
                 select
                   1
```

missing select 1? https://github.com/supabase/postgres-meta/pull/423/files#diff-036b3b055e43dc27784cbae9d673e7512eb49cefa7b7d74568bc85aad297a59bL36

sweatybridge (approved):

src/lib/PostgresMetaTypes.ts
soedirgo: It seems to be optional inside `exists`

PR #425, opened 2022-11-26
supabase/postgres-meta

Murkrage:


Murkrage:
Oh damn I just noticed Github Desktop decided to add a whole bunch of styling changes... I'll try and fix it 😅

soedirgo:
Hmm, `object` wouldn't really fit the bill here since it's too permissive (and tbh the type name doesn't help):

```ts
let x: object = [] as unknown[] // no type error
let y: object = (() => {}) as () => void // no type error
```

I believe your use case can be fulfilled by either:
- converting `TInstruction` into a type instead of an interface,
- add an [index signature](https://www.typescriptlang.org/docs/handbook/2/objects.html#index-signatures) to `TInstruction`: `[key: string]: Json`,
- or cast the `TInstruction` value to `Json` when assigning it

Murkrage:
Thanks for the reply @soedirgo! For whatever reason I hadn't considered the terrible type checking for `object`. Which makes sense because most things are an object underneath.

Converting to a `type` instead of an `interface` indeed did the trick. A little confused as to why as you should be able to use both from a TS perspective. All good now, though. Just need a minor refactor on my end but that's alright.

Closing this one :)

PR #442, opened 2022-12-23
chore: switch build to esnext module type

sweatybridge:

build

commonjs

## What is the new behavior?

no need to support commonjs as pg-meta is purely server side

## Additional context


inian (approved):

soedirgo:
FYI Studio imports @supabase/postgres-meta for types - not sure how this will affect that

PR #455, opened 2023-01-10
fix: #752 mark optional function parameters for typescript gen

Davepar:

Fix for #752, generating typescript types for a Postgres function with optional parameters.

A Postgres functions with optional parameters will output as required parameters, e.g. `p_user_id: string`.

## What is the new behavior?

The typescript output will mark optional parameters as optional, e.g. `p_user_id?: string`.

## Additional context

See bug for repro setup.


Davepar:
I tested this on several combinations of parameters:
* no parameters
* some required, some optional
* all optional

The other thing I considered is that the two variables I used, `pronargs` and `pronargdefaults`, are supported back as far as Postgres 9.

sweatybridge (approved):
LGTM. Thanks for contributing!

PR #469, opened 2023-01-17
feat: add typings for functions returning table

### LINKED ISSUE #468: Typings for functions returning setof <table>
# Feature request

## Is your feature request related to a problem? Please describe.

Currently, functions returning `setof <table>` aren't properly typed, e.g. returns `unknown`.

I noticed that `/functions` endpoint already resolves the `return_type` properly to `<table-name>` so I'm not sure if this is actually a bug or just not implemented yet. As the tables types will already be generated anyways I imagine it should be "easy" to copy them over?

I wonder if this might be intentional, but I can't think of any reason why it would be. However, I don't want to rule out the possibility that I'm missing something here, and I'm curious about your opinion.

## Describe alternatives you've considered

Manually casting the type.

jsbrain:

Fixes #468

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -222,7 +222,7 @@ export interface Database {
                 ([fnName, fns]) =>
                   `${JSON.stringify(fnName)}: ${fns
                     .map(
-                      ({ args, return_type }) => `{
+                      ({ args, return_type, complete_statement }) => `{
                   Args: ${(() => {
                     const inArgs = args.filter(({ mode }) => mode === 'in')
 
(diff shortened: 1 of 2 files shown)


soedirgo:
Hey, thanks! There's actually a bug in `return_type` - I'll fix that and then come back to this PR.

jsbrain:
Yeah I also realized meanwhile that my code doesn't work reliably. If you have 2 tables with the same name in 2 different schemas you would have the chance to get the typings of the wrong table in the "other" schema.

For this to work properly there would have to be more information in the `return_type`, like `public.users` instead of `users` or there might be another way to infer this correctly but I'm not sure.

soedirgo:
Yep, the plan is to add a `return_type_id` field, so we can disambiguate which table is referred exactly.

jsbrain:
Sounds like a plan 👍🏻

soedirgo:
Closing in favor of https://github.com/supabase/postgres-meta/pull/500

PR #477, opened 2023-01-20
Add HEALTHCHECK to Dockerfile

ruggi99:

Feature. Add `HEALTHCHECK` to `Dockerfile`
1. There is `/health` endpoint
2. Why not?

No issue

## What is the new behavior?

## Additional context

Set interval and timeout to 5 seconds, is it ok?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,3 +15,5 @@ COPY package.json ./
 ENV PG_META_PORT=8080
 CMD ["npm", "run", "start"]
 EXPOSE 8080
+# --start-period defaults to 0s, but can't be set to 0s (to be explicit) by now
+HEALTHCHECK --interval=5s --timeout=5s --retries=3 CMD curl -f http://localhost:8080/health || exit 1
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,4 +16,4 @@ ENV PG_META_PORT=8080
 CMD ["npm", "run", "start"]
 EXPOSE 8080
 # --start-period defaults to 0s, but can't be set to 0s (to be explicit) by now
-HEALTHCHECK --interval=5s --timeout=5s --retries=3 CMD curl -f http://localhost:8080/health || exit 1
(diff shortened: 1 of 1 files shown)


soedirgo (approved):
Thanks!

ruggi99:
We can add a Github Action to build, run and check health with Docker. What about it?

Dockerfile
```diff
@@ -15,3 +15,5 @@ COPY package.json ./
 ENV PG_META_PORT=8080
 CMD ["npm", "run", "start"]
 EXPOSE 8080
+# --start-period defaults to 0s, but can't be set to 0s (to be explicit) by now
+HEALTHCHECK --interval=5s --timeout=5s --retries=3 CMD curl -f http://localhost:8080/health || exit 1
```
sweatybridge, line 19: Are you sure curl exists in this docker image?

```bash
root@c4341b754f88:/usr/src/app# curl --version
bash: curl: command not found
```

soedirgo: Good catch - let's use this command instead: https://github.com/supabase/cli/blob/ab821f164ea0bbe88d27c4e1e39f024467c2cbd7/internal/start/start.go#L403

soedirgo:
> We can add a Github Action to build, run and check health with Docker

👍

Dockerfile
sweatybridge: If you don't want to install curl as a runtime dependency, an alternative is to use [node](https://github.com/supabase/cli/blob/main/internal/start/start.go#L403) for health check.

```suggestion
HEALTHCHECK --interval=5s --timeout=5s --retries=3 CMD node -e "require('http').get('http://localhost:8080/health', (r) => {if (r.statusCode !== 200) throw new Error(r.statusCode)})"
```

ruggi99: Good catch, didn't notice it was slim.
Agree with node version of the healthcheck cmd

ruggi99:
Sorry for not having tested it. Should work now.
Will PR a github action in order to skip this issues in the future

ruggi99:
@soedirgo need again an approval for workflow run I think

soedirgo (approved):

PR #490, opened 2023-01-28
fix: Fix logic when returning default roles

ruggi99:

Bug fix.
Fixes logic with `includeDefaultRoles`. Actually a breaking change.
Maybe I should change `includeDefaultRoles` to `true` to not change the behavior?

## What is the new behavior?

## Additional context

### DIFF
--- a/src/lib/PostgresMetaRoles.ts
+++ b/src/lib/PostgresMetaRoles.ts
@@ -1,5 +1,4 @@
 import { ident, literal } from 'pg-format'
-import { DEFAULT_ROLES } from './constants.js'
 import { rolesSql } from './sql/index.js'
 import {
   PostgresMetaResult,
@@ -43,7 +42,15 @@ FROM
 WHERE
   true`
     if (!includeDefaultRoles) {
-      sql += ` AND name NOT IN (${DEFAULT_ROLES.map(literal).join(',')})`
+      // All default/predefined roles start with pg_: https://www.postgresql.org/docs/15/predefined-roles.html
+      // The pg_ prefix is also reserved:
+      //
+      // ```
+      // postgres=# create role pg_mytmp;
+      // ERROR:  role name "pg_mytmp" is reserved
+      // DETAIL:  Role names starting with "pg_" are reserved.
+      // ```
+      sql += ` AND NOT pg_catalog.starts_with(name, 'pg_')`
     }
     if (limit) {
       sql += ` LIMIT ${limit}`
--- a/src/lib/PostgresMetaRoles.ts
+++ b/src/lib/PostgresMetaRoles.ts
@@ -42,7 +42,7 @@ FROM
   roles
 WHERE
   true`
-    if (includeDefaultRoles) {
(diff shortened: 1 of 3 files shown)


soedirgo:
The list of default/predefined roles can be checked [here](https://www.postgresql.org/docs/current/predefined-roles.html) - I don't think we need to expose this in the response.

src/lib/PostgresMetaRoles.ts
```diff
@@ -51,10 +51,12 @@ WHERE
...
       sql += ` OFFSET ${offset}`
     }
-    const result = await this.query(sql)
+    const result = (await this.query(sql)) as PostgresMetaResult<PostgresRole[]>
     if (result.data) {
-      result.data = result.data.map((role: any) => {
+      result.data = result.data.map((role) => {
+        // @ts-ignore
         role.config = changeRoleConfig2Object(role.config)
+        role.is_default_role = role.name in DEFAULT_ROLES
```
kevcodez, line 59: I believe you need to use `includes` here.

DEFAULT_ROLES is a string array

Screenshot 2023-01-28 at 12 06 13

kevcodez (dismissed):

src/lib/PostgresMetaRoles.ts
kevcodez: `in` would work for checking whether an object has a property

ruggi99:
Mmm yeah, I messed up a bit I think.
My proposed solution should add that field and tell Supabase interface to not edit that role because it's needed in Supabase ecosystem.
So for example, when and if Custom roles will be implemented Supabase interface should know if a role is editable or not.
That's what I'm trying to do.
But I realize that `DEFAULT_ROLES` is not usable here.
Should I try creating a new constant `SUPBASE_ROLES`?

src/lib/PostgresMetaRoles.ts
ruggi99: Thanks. I always forget about this in Javascript.
Some confusion between Javascript and Python

soedirgo:
No worries! We want to minimize Supabase-isms in pg-meta, so other users can benefit from it too.

> tell Supabase interface to not edit that role

By Supabase interface, do you mean the dashboard at https://app.supabase.com? If so, we can do this on the dashboard side. E.g. if you go to the [Table Editor](https://app.supabase.com/project/_/editor) and click on the schema dropdown, you'll see that there are "protected schemas":

Screenshot 2023-01-28 at 7 35 59 PM

We can do something similar for user-defined roles - we just dub Supabase internal roles "protected roles".

ruggi99:
@soedirgo yes I meant Supabase Dashboard.
So you are suggesting to hardcode "protected" roles in Supabase Dashboard?
Since it's a config (protected roles can change in time) I thought it would be best to put them in config API or something similar so I chose `postgres-meta`.
Since you want to keep `postgres-meta` Supabase agnostic I'll revert the changes and keep the bug fix

soedirgo (approved):

PR #493, opened 2023-01-30
fix: wrong `types` file

soedirgo:
From https://github.com/supabase/postgres-meta/pull/481

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   ],
   "type": "module",
   "main": "dist/server/app.js",
-  "types": "dist/lib/types.d.ts",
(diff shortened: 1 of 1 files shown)


sweatybridge (approved):

sweatybridge:
Any symptom of this error? Vscode seems to be importing types correctly for me when used in supabase/supabase.
Screenshot 2023-01-30 at 3 32 10 PM

soedirgo:
Yeah, e.g. `PostgresMeta` should be exported but isn't, and `postgres*Schema` consts shouldn't be exported but is.

PR #501, opened 2023-02-01
chore: test typegen

soedirgo:
Begin testing the server, starting with typegen.


ruggi99:
From what I see you're splitting server into production and development.
If that's true then you can update Dockerfile to include only production dependencies into the final stage (actually also development dependencies are included due to `swagger` for example).
Something like this:
```Dockerfile
FROM node:16-bullseye as deps
WORKDIR /app
# Do `npm ci` separately so we can cache `node_modules`
# https://nodejs.org/en/docs/guides/nodejs-docker-webapp/
COPY package.json package-lock.json ./
RUN npm clean-install

FROM deps as prod-deps
RUN npm prune --production

FROM deps as build
COPY . .
RUN npm run build

FROM node:16-bullseye-slim
WORKDIR /usr/src/app
COPY --from=prod-deps /usr/src/app/node_modules node_modules
COPY --from=build /usr/src/app/dist dist
COPY package.json ./
ENV PG_META_PORT=8080
CMD ["npm", "run", "start"]
EXPOSE 8080
# --start-period defaults to 0s, but can't be set to 0s (to be explicit) by now
HEALTHCHECK --interval=5s --timeout=5s --retries=3 CMD node -e "require('http').get('http://localhost:8080/health', (r) => {if (r.statusCode !== 200) throw new Error(r.statusCode)})"

```

soedirgo:
The lib/server split is kinda weird right now - `dependencies` is for `lib` and `dependencies` + `devDependencies` are for `server`.

Ideally `lib` and `server` are split into two TS projects - that can be done as a follow up.

kiwicopple (approved):

PR #533, opened 2023-03-11
Add capability for typescript type gen to reference table row types

joeally:

Bug fix

In the situation where we have a view in which one of the columns has the row type of a table, `unknown` is generated for that column.

e.g

```sql
CREATE VIEW rows_with_user_view AS (
  SELECT *, users as user FROM rows
  INNER JOIN users on rows.user_id = user.id
);

```

Currently results in the following generation

```typescript
{
...
  public: {
    ...
    Views: {
      rows_with_user_view: {
        Row: {
          ...
          user: unknown
        }
      }
    }
    ...
  }
...
}
```

## What is the new behavior?

Instead types should be generated in the following form

```typescript
{
...
  public: {
    ...
    Views: {
      rows_with_user_view: {
        Row: {
          ...
          user: Database["public"]["Tables"]["users"]["Row"]
        }
      }
    }
    ...
  }
...
}
```


src/server/templates/typescript.ts
```diff
@@ -419,7 +435,15 @@ const pgTypeToTsType = (
...
-
+    const tableRowType = tables.find((table) => table.name === pgType)
+    if (tableRowType) {
+      if (schemas.some(({ name }) => name === tableRowType.schema)) {
+        return `Database[${JSON.stringify(tableRowType.schema)}]['Tables'][${JSON.stringify(
+          tableRowType.name
+        )}]['Row']`
+      }
+      return 'unknown'
+    }
```
joeally, line 446: If the type is yet to be found we search in the tables of the same schema. If we find it we return the row type for that table.

soedirgo (approved):
Thanks! Sorry this took a year to get to.

PR #556, opened 2023-03-29
feat: table definitions

alaister:

feature

## What is the new behaviour?

A new `/definitions/<id>` endpoint where a user can send a table / view `id`, and get back the definition of that entity as formatted SQL. For example:
```sql
create table
  public.example (
    id integer not null,
    created_at timestamp with time zone default now() not null,
    updated_at timestamp with time zone default now() not null,
    name text not null,
    short_name text not null,
    excited_name text generated always as (upper(name)) stored,
    variable_name character varying(255),
    post_id uuid,
    constraint example_name_check check ((char_length(name) <= 4))
  );

create sequence public.example_id_seq as integer start
with
  1 increment by 1 no minvalue no maxvalue cache 1;

alter sequence public.example_id_seq owned by public.example.id;

alter table only public.example
alter column id
set default nextval('public.example_id_seq'::regclass);

alter table only public.example
add constraint example_pkey primary key (id);

alter table only public.example
add constraint example_post_id_fkey foreign key (post_id) references public.posts (id) on delete cascade;
```

## Additional context

TODO: add pg_dump to the docker image


alaister:
Closed in favour of using https://github.com/MichaelDBA/pg_get_tabledef instead

PR #559, opened 2023-04-04
feat: support vector data type when generating typescript types

gregnr:
Adds pgvector's `vector` data type to TypeScript generation logic to produce a ~~`number[]`~~ `string` type.

Edit: `vector` actually produces a `string`, eg `"[1,2,3]"`.

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -405,6 +405,8 @@ const pgTypeToTsType = (
     return 'undefined'
   } else if (pgType === 'record') {
     return 'Record<string, unknown>'
+  } else if (pgType === 'vector') {
+    return 'number[]'
   } else if (pgType.startsWith('_')) {
     return `(${pgTypeToTsType(pgType.substring(1), types, schemas)})[]`
   } else {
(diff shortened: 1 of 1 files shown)


soedirgo:
Thanks! I'm actually getting `vector` columns as `text` from PostgREST though:

```
create table t (c vector(3));
insert into t values ('[1,2,3]'), ('[4,5,6]');
```

```
[
  {
    "c": "[1,2,3]"
  },
  {
    "c": "[4,5,6]"
  }
]
```

soedirgo:
Ah seems like I can do:

```
> curl 'https://<...>.supabase.co/rest/v1/t?select=c::_float4' -H 'Authorization: Bearer <...>' -H 'apikey: <...>' | jq

[
  {
    "c": [
      1,
      2,
      3
    ]
  },
  {
    "c": [
      4,
      5,
      6
    ]
  }
]
```

This needs casting support on postgrest-js though - will take a look at that

gregnr:
Woah, TIL you can cast over PostgREST 😃

> This needs casting support on postgrest-js though - will take a look at that

Thanks! I imagine this means we should default it as a string instead (we probably shouldn't implicitly cast)?

soedirgo:
> we should default it as a string instead (we probably shouldn't implicitly cast)?

Yup, agreed :+1:

gregnr:
Great, updated it to a `string` now.

soedirgo (approved):
Thanks again!

PR #573, opened 2023-04-25
Typescript: safer nullability in functions, domain support

### LINKED ISSUE #842: Typescript gen : invalid SQL Function types
# Bug report

## Describe the bug

For both input and input types, typescript generation of SQL functions is invalid in some cases :
- Primitive types are generated as non-nullable (`type` instead of `type | null`) even though the function may accept NULL as input or output
- Domain types are generated as `unknown`

I think that this bug is not caused by a PostgreSQL limitation as views produce the correct typing : primitives are nullable, domains generate as their underlying type and when domains are marked `NOT NULL` they are even non-nullable in typescript.

## To Reproduce

```sql
CREATE DOMAIN strict_text AS text NOT NULL;

CREATE FUNCTION some_function(arg strict_text)
RETURNS table (nulltext text, stricttext strict_text)
LANGUAGE SQL AS $$
   SELECT NULL::uuid, arg
$$;
```

Generated type with `supabase gen types typescript --local`
```typescript
export interface Database {
  public: {
    Functions: {
      some_function: {
        Args: {
          arg: unknown
        }
        Returns: {
          nulltext: string
          stricttext: unknown
        }[]
      }
    }
  }
}
```

## Expected behavior

Generated types should be
```typescript
export interface Database {
  public: {
    Functions: {
      some_function: {
        Args: {
          arg: string
        }
        Returns: {
          nulltext: string | null
          stricttext: string
        }[]
      }
    }
  }
}
```

## System information

- OS: [e.g. Linux NixOS]
- Version of supabase CLI: 1.50.8
- Version of Node.js: 18.15.0

## Additional context

I :heart: the work being done and the mindset at Supabase. I believe strict typing is crucial to the success of a "backend in the DB" and I could help with a PR if I'm being given some pointers to the code responsible of the TS generation :wave:

ngasull:

Fixes https://github.com/supabase/postgres-meta/issues/842

```sql
CREATE DOMAIN strict_text AS text NOT NULL;

CREATE FUNCTION some_function(arg strict_text)
RETURNS table (nulltext text, stricttext strict_text)
LANGUAGE SQL AS $$
   SELECT NULL::text, arg
$$;
```

Generated type with `supabase gen types typescript --local`
```typescript
export interface Database {
  public: {
    Functions: {
      some_function: {
        Args: {
          arg: unknown
        }
        Returns: {
          nulltext: string
          stricttext: unknown
        }[]
      }
    }
  }
}
```

## What is the new behavior?

Generated types should be:
```typescript
export interface Database {
  public: {
    Functions: {
      some_function: {
        Args: {
          arg: string
        }
        Returns: {
          nulltext: string | null
          stricttext: string
        }[]
      }
    }
  }
}
```

Please review tests for more details on new expected behavior.

- Stop accepting non-nullable types in generated function arguments and return types.
- Add support for domains and take their nullability into account.
- Centralize nullability and array logic in `pgTypeToTsType`.

## Additional context

This is obviously a **breaking change** for most users using functions. It is however the safe way, exposing sound types.

### DIFF
--- a/src/lib/sql/types.sql
+++ b/src/lib/sql/types.sql
@@ -3,6 +3,8 @@ select
   t.typname as name,
   n.nspname as schema,
   format_type (t.oid, null) as format,
+  nullif(t.typbasetype, 0) as base_type_id,
+  not (t.typnotnull) as is_nullable,
   coalesce(t_enums.enums, '[]') as enums,
   coalesce(t_attributes.attributes, '[]') as attributes,
   obj_description (t.oid, 'pg_type') as comment
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -357,6 +357,8 @@ export const postgresTypeSchema = Type.Object({
   name: Type.String(),
   schema: Type.String(),
   format: Type.String(),
+  base_type_id: Type.Optional(Type.Integer()),
+  is_nullable: Type.Boolean(),
   enums: Type.Array(Type.String()),
   attributes: Type.Array(
     Type.Object({
--- a/src/server/routes/generators/typescript.ts
+++ b/src/server/routes/generators/typescript.ts
@@ -77,8 +77,7 @@ export default async (fastify: FastifyInstance) => {
       functions: functions.filter(
         ({ return_type }) => !['trigger', 'event_trigger'].includes(return_type)
       ),
-      types: types.filter(({ name }) => name[0] !== '_'),
-      arrayTypes: types.filter(({ name }) => name[0] === '_'),
+      types,
     })
   })
 }
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -80,8 +80,7 @@ if (EXPORT_DOCS) {
       functions: functions.filter(
         ({ return_type }) => !['trigger', 'event_trigger'].includes(return_type)
       ),
-      types: types.filter(({ name }) => name[0] !== '_'),
-      arrayTypes: types.filter(({ name }) => name[0] === '_'),
+      types,
     })
   )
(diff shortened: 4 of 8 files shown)


jumski:
Hey @ngasull Postgres domains have some troubles with **NOT NULL** constraint:

> CREATE DOMAIN in Postgres cannot safely enforce a NOT NULL constraint because domains are type aliases and do not directly validate constraints at the table level. While the domain itself can define a NOT NULL constraint, the database schema still allows null values in columns of that domain type unless the column explicitly defines NOT NULL. This behavior creates inconsistency: a domain's NOT NULL constraint is ignored unless explicitly reinforced on each column, leading to potential silent failures in enforcing the constraint.

But there is definitely a problem with serializing domain types with "gen-types" - they default to undefined.

ngasull:
@jumski This PR only aims to improve **function** generated types, no impact on tables. So the changes are still relevant.

Domains add a layer of safety over types at PostgreSQL runtime. Primitive types are unsafe (nullable) and it makes sense generate them as nullable unless they are a domain defined as NOT NULL.

Donnerstagnacht:
Hi @soedirgo , just a quick ping to see if you (or another maintainer) might have a chance to take a look at this PR when you have time. It touches on issues discussed in #841 and #842 and it is open since 2 years but still relevant.

eugeneford:
It would be incredibly amazing if we could have this done. When could we have this available?

SkeltalFlamingo:
We need this change! there's so much potential for runtime errors with inaccurate typing

avallete:

PR #585, opened 2023-06-06
fix(typegen): omit `columns` when querying tables

soedirgo:
Or views etc. This can be expensive on a complex enough schema and can easily trip over the 15s timeout.

Instead, query all columns separately, which should hopefully be less taxing and easier to plan.

Also omit excluded schemas when listing tables/views/etc.

Benchmark results on 100 tables with 100 columns each:

Before:
```
  Time (mean ± σ):      8.134 s ±  0.211 s    [User: 0.094 s, System: 0.017 s]
  Range (min … max):    7.857 s …  8.569 s    10 runs
```

After:
```
  Time (mean ± σ):      1.787 s ±  0.113 s    [User: 0.094 s, System: 0.016 s]
  Range (min … max):    1.648 s …  1.980 s    10 runs
```

### DIFF
--- a/src/server/routes/generators/typescript.ts
+++ b/src/server/routes/generators/typescript.ts
@@ -20,12 +20,31 @@ export default async (fastify: FastifyInstance) => {
 
     const pgMeta: PostgresMeta = new PostgresMeta({ ...DEFAULT_POOL_CONFIG, connectionString })
     const { data: schemas, error: schemasError } = await pgMeta.schemas.list()
-    const { data: tables, error: tablesError } = await pgMeta.tables.list()
-    const { data: views, error: viewsError } = await pgMeta.views.list()
+    const { data: tables, error: tablesError } = await pgMeta.tables.list({
+      includedSchemas: includedSchemas.length > 0 ? includedSchemas : undefined,
+      excludedSchemas,
+      includeColumns: false,
+    })
+    const { data: views, error: viewsError } = await pgMeta.views.list({
+      includedSchemas: includedSchemas.length > 0 ? includedSchemas : undefined,
+      excludedSchemas,
+      includeColumns: false,
+    })
     const { data: materializedViews, error: materializedViewsError } =
-      await pgMeta.materializedViews.list({ includeColumns: true })
+      await pgMeta.materializedViews.list({
+        includedSchemas: includedSchemas.length > 0 ? includedSchemas : undefined,
+        excludedSchemas,
+        includeColumns: false,
+      })
+    const { data: columns, error: columnsError } = await pgMeta.columns.list({
+      includedSchemas: includedSchemas.length > 0 ? includedSchemas : undefined,
+      excludedSchemas,
+    })
     const { data: relationships, error: relationshipsError } = await pgMeta.relationships.list()
-    const { data: functions, error: functionsError } = await pgMeta.functions.list()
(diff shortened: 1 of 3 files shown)


src/server/routes/generators/typescript.ts
```diff
@@ -55,6 +74,11 @@ export default async (fastify: FastifyInstance) => {
       reply.code(500)
       return { error: materializedViewsError.message }
     }
+    if (columnsError) {
+      request.log.error({ error: columns, request: extractRequestForLogging(request) })
```
kevcodez, line 78: error: columns => error: columnsError?

```suggestion
      request.log.error({ error: columnsError, request: extractRequestForLogging(request) })
```

src/server/templates/typescript.ts
```diff
@@ -78,8 +84,8 @@ export interface Database {
                     (table) => `${JSON.stringify(table.name)}: {
                   Row: {
                     ${[
-                      ...table.columns
-                        .sort(({ name: a }, { name: b }) => a.localeCompare(b))
+                      ...schemaColumns
+                        .filter((column) => column.table_id === table.id)
```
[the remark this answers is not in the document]
soedirgo: Yeah, makes sense

src/server/routes/generators/typescript.ts
soedirgo: Good catch!

src/server/templates/typescript.ts
```diff
@@ -345,8 +351,8 @@ export interface Database {
                     )
                     if (relation) {
                       return `{
-                        ${relation.columns
-                          .sort(({ name: a }, { name: b }) => a.localeCompare(b))
+                        ${schemaColumns
```
kevcodez, line 354: Just a sanity check, before you were passing different columns (views.columns, relation.columns) - it's correct that it's `schemaColumns` in all cases now as you're fetching them separately, right?

src/server/templates/typescript.ts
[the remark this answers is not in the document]
kevcodez: In case there are a lot of columns, filtering for every table may take a pretty long time (been there) - so it might make sense to map columns by table once, and then just use the table id as key for access.

This significantly improves speed with lots of tables + columns (it's crazy how slow these filters inside loops can get)

src/server/templates/typescript.ts
soedirgo: Yup, in this case we need to fetch it from `schemaColumns`

src/server/templates/typescript.ts
```diff
@@ -178,56 +183,42 @@ export interface Database {
...
-                      .sort(({ name: a }, { name: b }) => a.localeCompare(b))
-                      .map(
-                        (column) =>
-                          `${JSON.stringify(column.name)}: ${pgTypeToTsType(
-                            column.format,
-                            types,
-                            schemas
-                          )} ${column.is_nullable ? '| null' : ''}`
-                      )}
+                    ${columnsByTableId[view.id].map(
```
kevcodez, line 186: The columns were grouped by table id and you're accessing them via `view.id` here - are you sure you don't need another group by view id for these?

kevcodez (approved):
Assuming you tested this thoroughly, can't judge the functionality, but the code LGTM! And the performance improvements...

PR #586, opened 2023-06-07
fix: replace CTEs with joins

soedirgo:
Benchmark on 1000 tables, 4 columns each:

Schema:
```
do $$
    DECLARE myvar integer;
begin
    for myvar in 1..1000 loop
        EXECUTE format('CREATE TABLE sbtest%s (
        id SERIAL NOT NULL,
        k INTEGER NOT NULL,
        c CHAR(120) NOT NULL,
        pad CHAR(60) NOT NULL,
        PRIMARY KEY (id))', myvar);
    end loop;
end; $$
```

Workload: `GET /tables?include_columns=false`

Before: https://explain.dalibo.com/plan/h044ad2ae694gfc7
```
  Time (mean ± σ):      8.886 s ±  0.220 s    [User: 0.100 s, System: 0.022 s]
  Range (min … max):    8.666 s …  9.201 s    10 runs
```

After: https://explain.dalibo.com/plan/0a506f9b827e1b4c
```
  Time (mean ± σ):     404.1 ms ± 103.5 ms    [User: 94.7 ms, System: 15.1 ms]
  Range (min … max):   352.2 ms … 697.2 ms    10 runs
```

Conclusion: 20x speedup by using `group by` instead of [aggregate filters](https://www.postgresql.org/docs/current/sql-expressions.html#id-1.5.3.6.15.2).


src/lib/sql/tables.sql
```diff
@@ -16,10 +16,64 @@ SELECT
...
+      pg_index i,
+      pg_class c,
+      pg_attribute a,
+      pg_namespace n
+    where
+      i.indrelid = c.oid
+      and c.relnamespace = n.oid
+      and a.attrelid = c.oid
+      and a.attnum = any (i.indkey)
+      and i.indisprimary
```
soedirgo, line 47: This is just `primary_keys.sql` but inlined

test/lib/tables.ts
```diff
@@ -465,12 +465,12 @@ test('primary keys', async () => {
...
         "primary_keys": [
           {
-            "name": "c",
+            "name": "cc",
             "schema": "public",
             "table_name": "t",
           },
           {
-            "name": "cc",
+            "name": "c",
```
soedirgo, line 473: We don't guarantee ordering of pkeys, so 🤷

kevcodez (approved):
Code changes LGTM - can't really judge the functionality though

phamhieu (approved):

PR #596, opened 2023-07-20
feat: add type def to docs for column endpoints

owaisahmad-dev:

This PR adds types definitions to the Open API spec for the column CRUD endpoints.

Currently, the Open API spec doesn't contain the expected types for payloads and responses for column endpoints.

## What is the new behavior?

`npm run docs: export` now creates openapi.json which contains all type definitions for payloads and responses of column endpoints.
**Impacted endpoints:**
```
- GET /columns
- POST /columns
- GET /columns/{tableId}{ordinalPosition}
- PATCH /columns/{id}
- DELETE /columns/{id}
```

## Additional context

I was using these endpoints and wanted to generate a client using `openapi-generator-cli` and found that types are not hooked up for these endpoints. I took inspiration from the table endpoints and hooked it up like that. I am a newbie in open-source dev and would love any sort of feedback to improve.

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -47,7 +47,7 @@ export const postgresColumnCreateSchema = Type.Object({
   table_id: Type.Integer(),
   name: Type.String(),
   type: Type.String(),
-  default_value: Type.Optional(Type.String()),
+  default_value: Type.Optional(Type.Unknown()),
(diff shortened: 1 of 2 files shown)


soedirgo (approved):
Sorry for the delay - and thanks for the PR!

PR #600, opened 2023-08-03
feat: allow setting ssl root cert

### LINKED ISSUE #274: Error: [500] self signed certificate
# Bug report

I am trying to run Supabase on a Kubernetes cluster (with the [supabase-community/supabase-kubernetes](https://github.com/supabase-community/supabase-kubernetes) project). However, when connecting to the database I get the following error:

Error: [500] self signed certificate

image

As I use [zalando/postgres-operator](https://github.com/zalando/postgres-operator) to run the cluster, the TLS certificates get generated automatically. However, I am going to alter this behaviour and create custom certificates with cert-manager (`server.crt` and `ca.crt`). 

Where to I have to place them inside the postgres-meta container to make use of these files? What is the default path?

According to the `package.json` dependencies postgres-meta uses [pg](https://www.npmjs.com/package/pg), but I don't see a [default path for certificates](https://node-postgres.com/features/ssl) in the documentation.

soedirgo:

Feature

`sslmode` can be set, but `sslrootcert` can't be specified

## What is the new behavior?

Allow setting `sslrootcert`.

## Additional context

Fixes #274


test/index.test.ts
```diff
@@ -20,3 +20,4 @@ import './server/column-privileges'
 import './server/materialized-views'
 import './server/table-privileges'
 import './server/typegen'
+// import './server/ssl'
```
phamhieu, line 23: typo ?

phamhieu (approved):

test/index.test.ts
soedirgo: Intentionally omitted, since setting up the ssl key/certs was a pain. But I'll attempt it one last time

PR #605, opened 2023-08-25
feat: allow setting ssl root cert directly

soedirgo:
Set the root cert directly via env instead of setting the path of the cert. Also, `PG_META_DB_SSL_ROOT_CERT` will be applied on all pg connections, instead of the user having to set `&sslrootcert=` path in the connection string.

Follow up to https://github.com/supabase/postgres-meta/pull/600

### DIFF
--- a/src/server/constants.ts
+++ b/src/server/constants.ts
@@ -1,3 +1,5 @@
+import crypto from 'crypto'
+import { PoolConfig } from 'pg'
 import { getSecret } from '../lib/secrets.js'
 
 export const PG_META_HOST = process.env.PG_META_HOST || '0.0.0.0'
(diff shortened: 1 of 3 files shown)


src/server/constants.ts
```diff
@@ -23,17 +24,24 @@ if (!PG_CONNECTION) {
   pgConn.password = PG_META_DB_PASSWORD
   pgConn.pathname = encodeURIComponent(PG_META_DB_NAME)
   pgConn.searchParams.set('sslmode', PG_META_DB_SSL_MODE)
-  if (PG_META_DB_SSL_ROOT_CERT_PATH) {
-    pgConn.searchParams.set('sslrootcert', PG_META_DB_SSL_ROOT_CERT_PATH)
-  }
   PG_CONNECTION = `${pgConn}`
 }
 
+export const PG_META_DB_SSL_ROOT_CERT = process.env.PG_META_DB_SSL_ROOT_CERT
```
darora, line 30: Where is this getting picked up and used for the connections?

soedirgo: I keep getting brain farts... should be set now in DEFAULT_POOL_CONFIG

darora (approved):

PR #608, opened 2023-09-05
[WIP]add support to add multiple column to a table

mahendraHegde:

support to add multiple column to a table

we need to add columns 1 by 1

## What is the new behavior?

we will be able send columns as an array(batch) or object(backward compatibility) to the `POST /columns` endpoint

## Additional context

NA


mahendraHegde:
@soedirgo  PR is ready, plz review

PR #613, opened 2023-09-17
feat: bump node to 18

soedirgo:
Node 16 has been EOL'd as of 2023-09-11

### DIFF
--- a/.github/workflows/publish-deps.yml
+++ b/.github/workflows/publish-deps.yml
@@ -5,8 +5,8 @@ on:
 
 jobs:
   publish:
-    # Must match glibc verison in node:16-bullseye
-    runs-on: ubuntu-20.04
+    # Must match glibc verison in node:18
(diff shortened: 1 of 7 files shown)


.github/workflows/publish-deps.yml
```diff
@@ -5,8 +5,8 @@ on:
 
 jobs:
   publish:
-    # Must match glibc verison in node:16-bullseye
-    runs-on: ubuntu-20.04
+    # Must match glibc verison in node:18
+    runs-on: ubuntu-22.04
```
soedirgo, line 9: Don't know how to test this, but:
- ubuntu-20.04 uses glibc 2.31
- ubuntu 22.04 uses glibc 2.35
- node:16 uses glibc 2.31
- node:18 uses glibc 2.36

So this is as close as it gets.

sweatybridge (approved):

PR #614, opened 2023-09-20
fix(typegen): allow composite type attributes to be null

arin-c:

Typescript typegen bug fix

[CompositeTypes do not allow null values](https://github.com/supabase/supabase/issues/17539)

## What is the new behavior?

Generated composite types have nullable attributes.


tom-at-pixel:
This is a pretty important gap in Supabase. Is there a way to get some eyes on this review? 😊

PR #624, opened 2023-10-23
fix: use base/composite/enum type for fn columns

gulfaraz:

This PR changes `gen:types:typescript` to generate types for [function columns](https://github.com/supabase/postgres-meta/blob/master/src/server/templates/typescript.ts#L100-L109) instead of `unknown`.

The current output is,
```
// todos
Row: {
    details: string | null
    id: number
    "user-id": number
    blurb: string | null
    blurb_varchar: unknown | null
    details_is_long: unknown | null
    details_length: unknown | null
}
```
## What is the new behavior?

```
// todos
Row: {
    details: string | null
    id: number
    "user-id": number
    blurb: string | null
    blurb_varchar: string | null
    details_is_long: boolean | null
    details_length: number | null
}
```

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -99,14 +99,14 @@ export interface Database {
                       ),
                       ...schemaFunctions
                         .filter((fn) => fn.argument_types === table.name)
-                        .map(
-                          (fn) =>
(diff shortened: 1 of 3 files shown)


soedirgo:
Thanks for the PR @gulfaraz!

soedirgo (approved):

PR #642, opened 2023-11-02
fix: add integer to the list of numeric types

psteinroe:

bug fix

the return type `integer` of a function is `unknown`

## What is the new behavior?

`integer` is correctly parsed as `number`

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -428,7 +428,7 @@ const pgTypeToTsType = (
 ): string => {
   if (pgType === 'bool') {
     return 'boolean'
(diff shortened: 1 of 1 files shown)


kevcodez (approved):

soedirgo:
I'm not able to reproduce this issue:

```sql
postgres=# create function f() returns integer language sql as 'select 1';
CREATE FUNCTION
```

```ts
...
      f: {
        Args: Record<PropertyKey, never>
        Returns: number
      }
...
```

Note that an `integer` return type would show up as `int4` in this case, they're equivalent: https://www.postgresql.org/docs/current/datatype-numeric.html#DATATYPE-INT

PR #646, opened 2023-11-16
chore: bump deps

soedirgo:


kevcodez (approved):
Assuming we have enough tests to ensure this didn't break the API 

(typebox releases have the tendency to introduce a bunch of breaking changes)

soedirgo:
Good point - have gone through the [changelog](https://github.com/sinclairzx81/typebox/tree/master/changelog), I think we're in the clear 🤞

PR #654, opened 2023-11-20
feat: add Table and Enum helper types to typegen.

thorwebdev:

Add Table and Enum helper types to TypeScript typegen.

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -413,10 +413,85 @@ export interface Database {
     })}
 }
 
-export type Tables<T extends keyof Database["public"]["Tables"]> =
-  Database["public"]["Tables"][T]["Row"];
-export type Enums<T extends keyof Database["public"]["Enums"]> =
-  Database["public"]["Enums"][T];
+export type Tables<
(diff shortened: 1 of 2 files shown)


test/server/typegen.ts
```diff
@@ -363,6 +363,11 @@ test('typegen', async () => {
         }
       }
     }
+
+    export type Tables<T extends keyof Database["public"]["Tables"]> =
```
thorwebdev, line 367: @soedirgo as mentioned, should we make this more generic? Otherwise we kind of need a helper for insert/updates also and for all different schemas? Wdyt?

soedirgo: updated the PR

src/server/templates/typescript.ts
```diff
@@ -411,7 +411,88 @@ export interface Database {
           }
         }`
     })}
-}`
+}
+
+export type Tables<
```
thorwebdev, line 416: @kiwicopple do you think we need a separate type for `Views`? Or fine to use `Tables` for `Views`?

soedirgo (approved):

PR #685, opened 2024-01-03
chore(typegen): Extract metadata generation to function

isaacharrisholt:
The metadata collection for types was previously in the TypeScript generator route. This PR pulls it out into a separate function, which should make it easier to create type generators for other languages in the future (I plan on implementing types for Go imminently).

The only change introduced by this PR is to exit early from the function if a fetch fails, rather than performing all the fetches then going though the errors one by one. Should mildly increase performance for the unhappy path, though it was more to make the code easier to follow.

Refactor

Behaviour not changed.

## What is the new behavior?

Behaviour not changed.


kamilogorek (approved):
LGTM

soedirgo (approved):

PR #687, opened 2024-01-10
feat(typegen): Go Type Gen

isaacharrisholt:

Feature

## What is the current behaviour?

No Go types

## What is the new behaviour?

This PR adds type generation for Go based on the proposal in [this discussion](https://github.com/orgs/supabase/discussions/20132).

There's also some cleanup in `server.ts` for local development, which a) crosses off a `TODO` comment and b) makes it simpler to extend with new type generators later.

## Additional context

This PR requires #685 to be merged first and currently contains the changes from that PR. If you're interested in the changes made by this one, I recommend looking at commits `b24e63d` onwards, which mostly touch these files:

- `*/go.ts`
- `server.ts`
- `routes.ts`
- `package.json`

## Still TODO:

- [x] Handle composite types
- [x] Handle range types
- [x] Tests

Let me know your feedback or if you'd like any changes :)


isaacharrisholt:
I've now done the above, so this should be ready to go

soedirgo (approved):

PR #705, opened 2024-02-01
chore: bump to prettier 3

kevcodez:
Prettier 3 made formatting asynchronous, the only thing I wasn't sure about is the reassigned PostgresMeta.format function, I couldn't find any usage

Given we expose this as library, `PostgresMeta.format` has a breaking change, going from non-async to async


soedirgo:
There were problems when using Prettier 3 with Jest - we'll probably need to switch to Vitest

kevcodez:
@soedirgo Jest does not support snapshots with prettier 2, thus the prettierPath change here

PR #706, opened 2024-02-01
chore: bump to node 20

kevcodez:
- Bump to current LTS Node 20
- Update actions


kevcodez:
Unsure why the docker health check is failing

soedirgo (approved):

Issue #716, opened 2024-02-15
Hardcoded metadata resulting in broken TypeScript type generation from the Supabase CLI


yingw787:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

For differing PostgreSQL schemas outside of schema `public`, the hardcoded Postgres metadata for schema `public` does not work, and creates a TypeScript type error.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Install Supabase v1.144.7 in a TypeScript repository.
1. Create a PostgreSQL schema:

```sql
--- Create a new schema named `openai`.
CREATE SCHEMA openai;

--- Set the search path to include the new schema.
SET
  search_path TO openai,
  public;

-- Grant privileges to all tables for various Supabase roles so that the
-- Supabase client instantiated with the public Supabase access token can fetch
-- data, and RLS policies are still enforced per user.
GRANT USAGE ON SCHEMA openai TO postgres,
anon,
authenticated,
service_role,
dashboard_user;

ALTER DEFAULT PRIVILEGES IN SCHEMA openai
GRANT ALL ON TABLES TO postgres,
anon,
authenticated,
service_role,
dashboard_user;

COMMENT ON SCHEMA openai IS 'Schema for storing third-party API calls with OpenAI. This is important for e.g. archiving full conversation messages with OpenAI, which does not have an interface for exporting this information yet since the assistants functionality came out in beta a few days ago as of 11/08/2023.'
```

1. Start Supabase using `npx supabase start`
1. Reset the Supabase database using `npx supabase db reset`
1. Generate the TypeScript types using command or similar: `npx supabase gen types typescript --local --schema openai > src/types/supabase-schema-openai-types.ts`

## Expected behavior

A TypeScript file describing all of the tables, enums, and otherwise TypeScript-migrated models that I can use in a TypeScript project.

However, instead I am receiving this result:

```typescript
export type Json =
  | string
  | number
  | boolean
  | null
  | { [key: string]: Json | undefined }
  | Json[]

export interface Database {
  openai: {
    Tables: {
      assistants: {
        Row: {
          assistant_id: string | null
          checksum: string | null
          created_at: string
          file_ids: Json | null
          id: number
          model: string | null
          name: string | null
[...]

PR #717, opened 2024-02-15
fix(typegen): Fix support for custom SQL schemas in TypeScript type generation.

yingw787:

[x] Patches PR #654 to support SQL schemas outside of "public". Cannot directly use keyword "schema" due to circular type constraint error in definition `PublicTableNameOrOptions`.
[x] Update type generation tests to reflect updated TypeScript types, with local run of `npm run test` confirming 96 tests passed.

Addresses GitHub issue #716.

When running:

```bash
PG_META_GENERATE_TYPES_INCLUDED_SCHEMAS=openai PG_META_DB_URL=postgresql://postgres:postgres@127.0.0.1:54322/postgres npm run gen:types:typescript
```

Against a Supabase instance where database is at `postgresql://postgres:postgres@127.0.0.1:54322/postgres` and has schema `openai` among others, types `Table`, `TablesInsert`, `TablesUpdate`, and `Enums` result in type errors because schema `public` is not defined.

## What is the new behavior?

Against a Supabase instance where database is at `postgresql://postgres:postgres@127.0.0.1:54322/postgres` and has schema `openai` among others, types `Table`, `TablesInsert`, `TablesUpdate`, and `Enums` do not result in type errors.

## Additional context

N/A.


test/server/typegen.ts
```diff
@@ -367,7 +367,8 @@ test('typegen', async () => {
 
     export type Tables<
       PublicTableNameOrOptions extends
-        | keyof (Database["public"]["Tables"] & Database["public"]["Views"])
+        | keyof (Database[keyof Database]["Tables"] &
+            Database[keyof Database]["Views"])
```
sweatybridge, line 371: Hmm I think the concept here is to constrain each table to a new type on its own, for eg. `PublicTableNameOrOptions` type corresponds to the `public` schema exclusively.

We might need to generate separate `export type Tables` blocks depending on the schema name. @soedirgo wdyt?

yingw787: Ah sounds good! I was looking at that approach but since I don't personally use the new Table types I didn't know what the context of including them was. Happy to make this change (mapping over each schema, creating the PascalCase Table types, and updating the typegen types) and push up later this evening.

yingw787: @sweatybridge @soedirgo just pushed up some changes in order to create schema-specific types for Tables, TablesInsert, TablesUpdate, and Enums. Would love feedback on the source and test cases (should I add an additional test case to cover schemas not named public, and having multiple schemas)?

sweatybridge (dismissed):
This looks good to me. Will wait for @soedirgo's confirmation.

test/server/typegen.ts
sweatybridge: After some thought, I think there's a backwards compatibility issue with this change. So I proposed a new approach in https://github.com/supabase/postgres-meta/pull/723 and added you as the co-author.

PR #723, opened 2024-02-20
fix(typegen): handle missing public schema

sweatybridge:

supersedes https://github.com/supabase/postgres-meta/pull/717

If public schema is not requested, the generated type fails to compile.

## What is the new behavior?

When `public` schema is not requested, `Tables<'name'>` is now typed as `unknown`.

To use any non-public schema, pass in explicitly to `Tables<{ schema: 'private' }, 'name'>`.

## Additional context

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -419,9 +419,11 @@ export type Database = {
     })}
 }
 
+type PublicSchema = Database[Extract<keyof Database, "public">]
+
 export type Tables<
   PublicTableNameOrOptions extends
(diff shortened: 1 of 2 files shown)


yingw787:
I just ran `PG_META_GENERATE_TYPES_INCLUDED_SCHEMAS=openai PG_META_DB_URL=postgresql://postgres:postgres@127.0.0.1:54322/postgres npm run gen:types:typescript | xclip -sel clip` and it works for me, so I'm happy with this!

soedirgo (approved):

PR #727, opened 2024-02-21
fix(typegen): handle tables without any columns

sweatybridge:

Bug fix https://github.com/supabase/postgres-meta/issues/726

## What is the new behavior?

Generating types for `create table empty()` should not crash.

## Additional context

### DIFF
--- a/test/server/typegen.ts
+++ b/test/server/typegen.ts
@@ -30,6 +30,12 @@ test('typegen', async () => {
             }
             Relationships: []
           }
+          empty: {
+            Row: {}
+            Insert: {}
+            Update: {}
+            Relationships: []
+          }
           memes: {
             Row: {
               category: number | null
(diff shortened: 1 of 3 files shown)


test/server/typegen.ts
```diff
@@ -30,6 +30,12 @@ test('typegen', async () => {
             }
             Relationships: []
           }
+          empty: {
+            Row: {}
+            Insert: {}
+            Update: {}
```
soedirgo, line 36: These should be `Record<PropertyKey, never>` instead of `{}`, [because TypeScript™️](https://www.reddit.com/r/typescript/comments/tq3m4f/the_difference_between_object_and_recordstring/)

soedirgo: Eh, probably not an issue, so nvm

soedirgo (approved):

PR #739, opened 2024-03-12
fix: rate limit strikes back

soedirgo:
GitHub rate limit, go pound sand


kamilogorek:
For the reference: https://github.com/semantic-release/github?tab=readme-ov-file#options

kamilogorek (approved):

PR #746, opened 2024-03-20
fix: triggers w/ the same name on different schemas are duplicated [GEN-8547]

### LINKED ISSUE #715: Duplicated triggers in one schema.
# Bug report

## Describe the bug

There is a bug in UI Side (not functionally) that duplicated triggers in one schema, though the triggers were created to different schemas.

## To Reproduce

1. Go to Supabase Dashboard.
2. Create 2 functions with same name, but pointed to 2 different schemas.
3. Create 2 triggers from the previous functions respectively.
4. Refresh the page.
5. Found that both functions classified as only of the schemas, in the UI.

## Expected behavior

There shouldn't be a bug that show duplicated trigger in the same schema, while actually they live in different schemas.

## Screenshots & Video

Screenshot 2023-01-15 at 00 41 35
Screenshot 2023-01-15 at 00 41 23

https://user-images.githubusercontent.com/66579441/212489886-256100da-5200-40b9-926c-39c2726399fa.mp4

## System information

- OS : MacOS
- Browser : Safari

soedirgo:
Closes #715

### DIFF
--- a/src/lib/sql/triggers.sql
+++ b/src/lib/sql/triggers.sql
@@ -29,6 +29,7 @@ ON pg_t.tgrelid = pg_c.oid
 JOIN information_schema.triggers AS is_t
 ON is_t.trigger_name = pg_t.tgname
 AND pg_c.relname = is_t.event_object_table
+AND pg_c.relnamespace = is_t.event_object_schema::regnamespace
 JOIN pg_proc AS pg_p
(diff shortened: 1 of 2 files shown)


kamilogorek (approved):
Love all those 1-line fixes with proper regression tests 🥰

PR #750, opened 2024-03-21
fix(typegen): typescript: deprecate `Json` type in favour of `unknown`

### LINKED ISSUE #676: Generate types: "Json" type is incompatible with other types
**Describe the bug**
When using the generate types feature for your Database (which is mostly great), if your column type is `json` or `jsonb`, the generated type it gives you is this:

```ts
export type Json = string | number | boolean | null | { [key: string]: Json | undefined } | Json[]
```
This type is incompatible with any type you try to use it for. For example, if I'm using a column to hold an ingredient (this is hypothetical):

```ts
export interface Ingredient {
  id: number
  name: string
  quantity: string
}
const result: Json = { id: 1, name: 'Berries', quantity: '1 cup' } // this is how the object will be typed using the generated types
const ingredient: Ingredient = result // Type '{ [key: string]: Json | undefined; }' is missing the following properties from type 'Ingredient': id, name, quantity(2739)
const ingredient2: Ingredient = result as Ingredient 
// above results in:
// Conversion of type '{ [key: string]: Json | undefined; }' to type 'Ingredient' may be a mistake because neither type sufficiently 
// overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```
I have more examples of this in the typescript playground below. Given how error-prone it is to add a type for JSON data, can you please consider using `any` for Json types? We have to cast these values as `any` anyway because you cannot cast the type you are using, as demonstrated above. 

**To Reproduce**
You can see this behavior on this TypeScript playground link:

https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcBSBnCA7OBeODGKAS2wHM4AfObAVwFsAjYKauJiCAG2AENcNet27sA3nADaAa2CIAXIWJlyAXUWYc7OtgAmwAGZlguuAF92m7JNUAoewHoHcAKIANAIIBZAAoAZFzgARjgAWjgIJgArYABjGAx7UEhYODIYVgM+WNQASQooExJgbHgxWzg03UV6ZlYK2j4GYEUiUgoGgEc6ARgSJFblDrNbZOh4dMzs1AAlOJIUOHLK7CaWpXbyAG4GlULdYtKMRXzyfcOYGx2R21icIjhCjDpuGAAGDSxcAgkSauCADSNZqKADkACFWKRgBhQUDur1+go4KCQrE6GBQeZbvcJgUiiUYCd8QdCfhHjCXu84E44AAVZCoUESGRyQabdToL7aPSGYy6LbmLEkDBwBgijAqBAAC1QBh43AgAHcpWAoBAULBiqKDOqGAhGSjTudCaDFH8gatmvCeqUkQAKABMAHYAMwATgAlDjsA89gTSo7iWcA-ACE8qW84HxRcbQ3BbLS+JwAG6oCOvUVkeSJ5wAYRwaagkq0EAMBsWzKksmRbRUnKsPP0RmwJkFZixMAgFaZcdJpSxDD4iA4qD4YpFMD4slHsT4dAwqFb-VlbCQi2eBiMsQu3BHuYiRe4fDAopVMGlMtQEAvrAAdHBcuWLyK4EqY2lSoSSDg+NwgXc2BFvAt5wMkTwlrgXYojo0jYMq2BYkYxYwHe9iAQ8GZElyWg-A0KxrGCADKDBcC+wBwvhn4hv2CSKJIVGVJUvz-EElqESikJQNCsI2oiAwomiGJYmYAKMUxLGKGxwLrKCXgkNw0hwnACJ2gJoKOnA6KniJYlMfpSxVFJ7EgiiAASOByMpql9OpIQwEwGCYuYelMXYNwYfAhQ7igihzD5qDhpSryCrSDKViyNbsvWnxaDQOjNvy7bCqK4oYJKFBXnA8oiMqqrqpqfQwtleo9ii-kLBRNRrEC-q0RgTpul6jjOO43j+IEmnhHw3HDuhuIUs8rxBjh3xSFRkmAjJYJcTx1m2rZyKolpwkuRNRlTVasnyYp838UtmnabCa36ZN0lbWCFmtoge1qUt9mOc5om2HYPp+iSFzBiapTklhjqSG8qg0s44VMlYKW0De0bpSQ5CrEwvAIN2669h9ppoQebW+AEcCumEGxStw-SsH+-W+l5wUwK6sVjaC8oQKC9go3AABiXDknTXCM298CgE0YC8IobPdkFQ1U8D9KGqCdYUBD8HwDGkpw8miPQczoIAET0xroIY7YQA

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 120
 - Version of CLI: 1.93.0
 - Version of supabase-js: 2.38.0
 - Version of Node.js: 20.5.1

isaacharrisholt:

Bug fix

#676

## What is the new behavior?

`Json` has been marked as deprecated and is now just equivalent to `unknown`.

FIxes #676

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -36,363 +36,373 @@ export type Json = unknown
 
 export type Database = {
   ${schemas
-      .sort(({ name: a }, { name: b }) => a.localeCompare(b))
-      .map((schema) => {
-        const schemaTables = tables
-          .filter((table) => table.schema === schema.name)
-          .sort(({ name: a }, { name: b }) => a.localeCompare(b))
-        const schemaViews = [...views, ...materializedViews]
-          .filter((view) => view.schema === schema.name)
-          .sort(({ name: a }, { name: b }) => a.localeCompare(b))
-        const schemaFunctions = functions
-          .filter((func) => {
-            if (func.schema !== schema.name) {
-              return false
-            }
-
-            // Either:
-            // 1. All input args are be named, or
-            // 2. There is only one input arg which is unnamed
-            const inArgs = func.args.filter(({ mode }) => ['in', 'inout', 'variadic'].includes(mode))
-
-            if (!inArgs.some(({ name }) => name === '')) {
-              return true
-            }
-
-            if (inArgs.length === 1) {
-              return true
-            }
-
+    .sort(({ name: a }, { name: b }) => a.localeCompare(b))
+    .map((schema) => {
+      const schemaTables = tables
+        .filter((table) => table.schema === schema.name)
+        .sort(({ name: a }, { name: b }) => a.localeCompare(b))
+      const schemaViews = [...views, ...materializedViews]
+        .filter((view) => view.schema === schema.name)
+        .sort(({ name: a }, { name: b }) => a.localeCompare(b))
+      const schemaFunctions = functions
+        .filter((func) => {
+          if (func.schema !== schema.name) {
             return false
-          })
-          .sort(({ name: a }, { name: b }) => a.localeCompare(b))
-        const schemaEnums = types
-          .filter((type) => type.schema === schema.name && type.enums.length > 0)
-          .sort(({ name: a }, { name: b }) => a.localeCompare(b))
-        const schemaCompositeTypes = types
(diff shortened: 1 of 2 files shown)


avallete:

PR #768, opened 2024-05-02
fix: query timeout

soedirgo:

Bug fix

Testing locally, `select pg_sleep(<forever>)` just hangs even past the `connectionTimeoutMillis`

## What is the new behavior?

Use `query_timeout` instead which returns `Query read timeout` with a 400

### DIFF
--- a/src/server/constants.ts
+++ b/src/server/constants.ts
@@ -43,7 +43,7 @@ export const GENERATE_TYPES_DETECT_ONE_TO_ONE_RELATIONSHIPS =
 
 export const DEFAULT_POOL_CONFIG: PoolConfig = {
   max: 1,
-  connectionTimeoutMillis: PG_CONN_TIMEOUT_SECS * 1000,
+  query_timeout: PG_CONN_TIMEOUT_SECS * 1000,
   ssl: PG_META_DB_SSL_ROOT_CERT ? { ca: PG_META_DB_SSL_ROOT_CERT } : undefined,
 }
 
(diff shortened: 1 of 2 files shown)


phamhieu (approved):

src/server/constants.ts
```diff
@@ -43,7 +43,7 @@ export const GENERATE_TYPES_DETECT_ONE_TO_ONE_RELATIONSHIPS =
 
 export const DEFAULT_POOL_CONFIG: PoolConfig = {
   max: 1,
-  connectionTimeoutMillis: PG_CONN_TIMEOUT_SECS * 1000,
+  query_timeout: PG_CONN_TIMEOUT_SECS * 1000,
```
darora, line 46: A connection timeout is *not* a query timeout. imo it's very misleading to feed in a "CONN_TIMEOUT" here.

src/server/constants.ts
```diff
@@ -43,7 +43,7 @@ export const GENERATE_TYPES_DETECT_ONE_TO_ONE_RELATIONSHIPS =
 
 export const DEFAULT_POOL_CONFIG: PoolConfig = {
   max: 1,
-  connectionTimeoutMillis: PG_CONN_TIMEOUT_SECS * 1000,
```
darora, line 46: Removing this results in the default of 0 taking over - which is **not** what we want.

src/server/constants.ts
soedirgo: Good catch, made a PR to revert it: https://github.com/supabase/postgres-meta/pull/777

PR #779, opened 2024-06-12
feat(typegen): add swift template

grdsdev:

Close https://github.com/supabase/supabase-swift/issues/240
Add a new generator for generating Swift types.

### v0

This is considered a v0 for the Swift types generation.

- Generated types are plain Swift structs, which can be used to serialize/deserialize Postgrest requests/responses
- Types are not integrated into the client libs in a way such as the JS library

### DIFF
--- a/src/server/templates/swift.ts
+++ b/src/server/templates/swift.ts
@@ -17,48 +17,56 @@ type SwiftGeneratorOptions = {
   accessControl: AccessControl
 }
 
-function generateEnum(enum_: PostgresType, options: SwiftGeneratorOptions): string {
-  return `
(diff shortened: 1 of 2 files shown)


src/server/templates/swift.ts
```diff
@@ -0,0 +1,283 @@
...
+    if (tableRowType) {
+      return `${toUpperCamelCase(tableRowType.name)}`
+    }
+
+    const viewRowType = views.find((view) => view.name === pgType)
+    if (viewRowType) {
+      return `${toUpperCamelCase(viewRowType.name)}`
+    }
+
+    return 'unknown'
```
grdsdev, line 238: TODO: return an `AnyJSON` type here

soedirgo (approved):

PR #789, opened 2024-07-01
fix: support gen types from command line

sweatybridge:

feature

## What is the new behavior?

Support typegen from cli mode so that scripts in [package.json](https://github.com/supabase/postgres-meta/blob/master/package.json#L26) continue to work.

## Additional context


grdsdev (approved):
thanks!

Issue #795, opened 2024-07-19
Generating types in python format


dhruvmalik007:
**Is your feature request related to a problem? Please describe.**

I was following the tutorial for generating the database types for supabase [here](https://www.restack.io/docs/supabase-knowledge-supabase-generate-types) , so I thought normally this command can be adapted for the python in order to generate the corresponding types in python objects

**Describe the solution you'd like**

to include the command 

`npx supabase gen types python  --project-id "$PROJECT_REF"   --schema public > types.py`  should generate the types in the python.

**Describe alternatives you've considered**
converting the types generated in ts into py ;) . 
**Additional context**
N.A

Mohmn:
Could you please assign this issue to me? I find it very interesting t. Also, could you guide me on anything I need to know to get started? Any specific instructions or resources would be greatly appreciated

dhruvmalik007:
thanks @Mohmn  for the response . my bad for late response but yeah I understood we will have to define the generator types and the wrapped call functions in order to pythonic format.

faroukcharkas:
python types would actually be incredible, having a built in almost-ORM? with supabase would be wild

urieltfk:
Please, this would be amazing.

antopiahk:
Please, this would be beyond amazing.

nikita-olechko:
Wrote a small implementation for Python here: 

https://github.com/nikita-olechko/Supabase-Table-Type-Definitions-In-Python/blob/main/README.md

Far from comprehensive, but works.

ryanpeach:
https://github.com/supabase/postgres-meta/pull/808

tresorama:
Any updates on this ?

ryanpeach:
If it could get testing guidance on my PR I'd continue my work.

tresorama:
> If it could get testing guidance on my PR I'd continue my work.

You pr would add a new command to supabase cli ?

ryanpeach:
No it just adds Python type generation.

tresorama:
> No it just adds Python type generation.

So it's a standalone script .py ?

ryanpeach:
... no. It's been a while but as I remember it the CLI generates types and has a language flag. I'm adding Python as a language.

ryanpeach:
[...]

PR #808, opened 2024-10-01
Python Language Generation

ryanpeach:

Feature

Does not generate python, 

## What is the new behavior?

Practicing TDD, designing tests first and asking the community for feedback.

## Additional Comments

I'd like everyones feedback on this format for the python types. I'm using pydantic and mimicing the go structure, since its closest to python dataclasses (basically structs).


ryanpeach:
I'm not really sure what a lot of these things other than "Tables" are. Materialized Views? Why are there different types for different operations? Composites?

ryanpeach:
I'd also like to do this again, but for rust.

lwih:
Man I'm looking forward to that :)

yangcheng:
happy to be early tester!

ryanpeach:
I've dropped my work on it atm if someone wants to take it the final mile, or wants to answer my questions on this pr, that would be great! Eventually i'll get back to working on supabase.

ryanpeach:
I need someone to check the tests. I’m know I’ve not covered all cases, but I don’t understand some of the test cases in the other code generators. Basically I need a code review.

mikelong10:
Hey, y'all 👋 just wanted to join the convo here

I'm working on a project with a Next.js frontend, Python FastAPI backend, and Supabase, and recently realized there was no  python type gen support :(

I came across #795 when I was Googling around and then saw this PR here. Hoping to spark the convo and get the ball rolling again here! I did a quick look through, and the changes seem on the right track to me.

I'm def a little out of my domain here, but I saw @sweatybridge you commented on the original #795 saying you'd be more than happy to review a PR for this issue. Seems like @ryanpeach would love to get your or someone else from Supabase's eyes on this!

grdsdev:
Hi @ryanpeach, thanks a bunch for all the hard work you put into this PR!

I’d love to chat about this and figure out how we want to approach Python type generation. The way you implemented it, as raw Python models, sounds great, but we might want to take it a step further and try something more aligned with what we have in JS.

In `postgrest-js`, the generated types are integrated into the `PostgrestClient` as you can see on https://github.com/supabase/postgrest-js/blob/master/src/PostgrestClient.ts#L18-L24.

This makes types available during runtime usage of the library.

I’m not a Python expert, so I’m not sure how far Python types can go.

ryanpeach:
@grdsdev I did see that, but I think we should go for language availability before we go for complexity. That's why I emulated the go structs. Maybe all that can go in a V2.

ryanpeach:
Either way, I think the best solution is this:

1. Create a canonical "schema" for the database in TDD
2. Create an agreed upon "test" for the language generation in a PR. Evaluate it and vote on it.
3. Write the code to generate the test. Release.

soedirgo:
> we should go for language availability before we go for complexity. That's why I emulated the go structs

I'll have to disagree on this - the typegen needs to be well integrated into and narrowly scoped into the use case of the SDK. I don't see how generating structs/interfaces on its own would improve the Supabase DX.

I'd actually use the Go types as a reason against adding this, because to this day supabase-go still doesn't use the Go types (cmiiw), and it's not clear how you're meant to use the two together.

soedirgo:
To elaborate further: some of the reasons we added typings to supabase-js are because it affords us:
- autocompletion (table name in `.from()`, column name in filters)
- static type checking for missing columns in `.select()`
- typed result based on the query in `.select()`

I'm not convinced of the benefits of adding typings if it doesn't afford us these things.

ryanpeach:
@soedirgo you aren't convinced that having structs which match your tables, and having those structs autogenerated, is useful?

I'm sorry but I can't even begin to see the logic in that perspective... It's a basic serialization/deserialization paradigm. It's 1000x better than having an untyped object without data validation, static type checking, or IDE autocompletion.

And I think the interest in this issue is clear that people care about it. How many of the tools developed here ad-hoc by developers to fix this issue just for python go beyond struct generation?

soedirgo:
I see where you're coming from; if Python's typings doesn't support these features then it is what it is. I'll leave it to the supabase-py maintainers to decide if we want to add this in, because if we do want to add these features in v2 they'd be responsible for integrating the types with supabase-py. cc @juancarlospaco @silentworks @grdsdev

soedirgo:
What would also help is if you could show an example of how you'd use these types in your code, because certain approaches end up making the code much more verbose and it gets unclear whether it's a DX improvement or not (example [here](https://github.com/supabase/supabase-flutter/issues/139))

ryanpeach:
I'd just use pydantics deserialization and from dict capabilities after running a query, just to associate a type with the return and get data validation.

silentworks:
Great work here @ryanpeach. Can you provide a code example of how this would be used in a Python project?

ryanpeach:
```
import * from generated_table_classes

response = PublicUsersSelect(supabase.table("planets").select("*").execute())
```

silentworks:
I think this would be a great addition to the CLI for sure.

mikelong10:
^ from testing locally, this seems to get the static type checking.

However, I will say that I have the "Mypy Type Checker" VSCode extension installed, and I don't believe by default Pydantic is able to catch the type errors of missing required fields in the `UsersInsert` class. For example, the Mypy extension picks this up, but if I disable the extension I don't get this red squiggly anymore:

image

Everything else seems good and working, though with intellisense picking up on it.

image
image
image
image

ryanpeach:
> Yeah I think this looks great! In terms of what's useful/valuable and would improve DX, I'd have to agree with @ryanpeach.
> 
> Essentially, even just simple auto-generated pydantic model classes that help you automatically stay up to date with your supabase schema and give you static type-checked, validated data to work with is enough of a DX improvement to make this effort worth it in my opinion 🙏
> 
> From your last example, maybe we could do this to make it even cleaner and clearer? Let me know your guys' thoughts and if you guys think something like this would be doable! cc: @silentworks
> ### Usage:
> 
> ```
> from generated_table_classes import Users, UsersInsert, UsersUpdate
> 
> insert_user_1 = Users.insert(UsersInsert(name="Ryan", email="ryan@gmail.com"))[0]
> insert_user_2 = Users.insert(UsersInsert(name="Mike"))[0]
> 
> select_user_1 = Users.select("*", filters={"id": insert_user_1.id})[0]
> select_all_users = Users.select("*")
> 
> update_user_1 = Users.update(
>     UsersUpdate(email="ryan@gmail.com"), filters={"id": insert_user_1.id}
> )[0]
> update_user_2 = Users.update(
>     UsersUpdate(name="Michael"), filters={"id": insert_user_2.id}
> )[0]
> ```
> 
> ### Implementation:
> 
> ```
> # generated_table_classes.py
> 
> # User schema, required name, optional email
> class User(BaseModel):
>     id: Annotated[int, Field(alias="id")]
>     name: Annotated[str, Field(alias="name")]
>     email: Annotated[str | None, Field(alias="email")] = None
> 
> 
> # UsersInsert schema, required name, optional email
> class UsersInsert(BaseModel):
>     name: Annotated[str, Field(alias="name")]
>     email: Annotated[str | None, Field(alias="email")] = None
> 
> 
> # UsersUpdate schema, optional name, optional email
> class UsersUpdate(BaseModel):
>     name: Annotated[str | None, Field(alias="name")] = None
>     email: Annotated[str | None, Field(alias="email")] = None
> 
> 
> # Single Users class used to interact with the users table
> class Users:
>     def __init__(self, users: list[dict]):
>         self.users = [User(**x) for x in users]
> 
>     def __iter__(self):
>         return iter(self.users)
> 
>     @staticmethod
>     def select(
>         *columns: str,
>         filters: Dict[str, Any] | None = None,
>         limit: int | None = None,
>         order_by: str | None = None,
>         descending: bool = True,
>     ) -> List[User]:
>         query = supabase.table("users").select(*columns)
> 
>         if filters:
>             for column, value in filters.items():
>                 if column == "or":
>                     query = query.or_(value)
>                 else:
>                     query = query.eq(column, value)
> 
>         if order_by:
>             query = query.order(order_by, desc=descending)
> 
>         if limit:
>             query = query.limit(limit)
> 
>         response = query.execute().data
>         return [User(**x) for x in response]
> 
>     @staticmethod
>     def insert(user: UsersInsert) -> List[User]:
>         response = supabase.table("users").insert(user.model_dump()).execute().data
>         return [User(**x) for x in response]
> 
>     @staticmethod
>     def update(
>         user: UsersUpdate,
>         filters: Dict[str, Any] | None = None,
>     ) -> List[User]:
>         query = supabase.table("users").update(user.model_dump())
>         if filters:
>             for column, value in filters.items():
>                 if column == "or":
>                     query = query.or_(value)
>                 else:
>                     query = query.eq(column, value)
> 
>         response = query.execute().data
>         return [User(**x) for x in response]
> ```

I think you need to keep it returning the Insert object and the Select object, otherwise you can't follow up with [filters](https://supabase.com/docs/reference/python/using-filters) and [modifiers](https://supabase.com/docs/reference/python/using-modifiers)

MitchellGRead:
curious if there is a reason this hasn't been merged yet?

soedirgo:
Just need to mark this PR as ready for review and a +1 from @silentworks, and then we can merge this.

sourman:
@ryanpeach thanks for the contribution! what help do you need to get this PR merged?

ryanpeach:
It's been a while, I will work on it again if I need a supabase project in the future, but I never got anyone to comment on the unit testing.

test/server/typegen.ts

If anyone officially with the project just would modify that test to make it look like the maintainers want it to look, then implementation is not difficult.

src/server/templates/python.ts
```diff
@@ -0,0 +1,304 @@
...
+    },
+    [0, 0]
+  )
+
+  // Pad the formatted name and type to align the struct fields, then join
+  // create the final string representation of the struct fields.
+  const formattedAttributeEntries = attributeEntries.map(([formattedName, type, name]) => {
+    return `  ${formattedName.padEnd(maxFormattedNameLength)} ${type.padEnd(
+      maxTypeLength
+    )} \`json:"${name}"\``
```
o-santi: Not correct. Seems to be related to golang struct generation?

src/server/templates/python.ts
```diff
@@ -0,0 +1,304 @@
...
+  tsrange: 'str',
+  tsmultirange: 'str',
+  tstzrange: 'str',
+  tstzmultirange: 'str',
+  daterange: 'str',
+  datemultirange: 'str',
+
+  // Miscellaneous types
+  void: 'None',
+  record: 'dict[str, Any]',
```
o-santi: Can't we infer more than just `Any`?

o-santi (changes requested):
I've left a small review with some problems. I'll start working on these problems, and also on getting the code cleaned up so that it can be merged.

avallete:
Thanks for this. Type generation has moved out of `postgres-meta`: since v0.99.0 this repo consumes [`@supabase/postgrest-typegen`](https://github.com/supabase/sdk/tree/main/packages/postgrest-typegen) and the templates this PR added to were deleted in #1084.

Python generation already lives in that package (the current maintainer template, not this 2023 draft). The generators there are frozen for now — they are being relocated into each language's own SDK repository. Closing here as the target code no longer exists.

PR #809, opened 2024-10-09
fix(typegen): don't generate relationships across schemas

soedirgo:
PostgREST doesn't detect relationships across schemas

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -151,6 +151,7 @@ export type Database = {
                       .filter(
                         (relationship) =>
                           relationship.schema === table.schema &&
+                          relationship.referenced_schema === table.schema &&
                           relationship.relation === table.name
                       )
                       .sort(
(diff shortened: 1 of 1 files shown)


avallete (approved):
LGTM, maybe we'll want this not only in typescript tough. So maybe some filter at relation fetching level 🤔 

Also, I wonder if that make sense for `postgres-meta` to not handle cross-schemas relations. PostgREST might not handle it, but postgres itself does.

soedirgo:
> maybe we'll want this not only in typescript tough. So maybe some filter at relation fetching level

IIRC these are used by the dashboard to detect fkeys across schemas, so we need to keep it as is.

> I wonder if that make sense for postgres-meta to not handle cross-schemas relations. PostgREST might not handle it, but postgres itself does.

The typegen is designed with PostgREST use case in mind, although it's not documented as such - we may want to make this explicit at some point though (e.g. rename the route to `postgrest-types/typescript`)

PR #821, opened 2024-11-05
fix: bump to fastify 5 and resolve vulnerabilities

kevcodez:


src/lib/Parser.ts
```diff
@@ -35,7 +33,7 @@ interface ParseReturnValues {
  */
 export function Deparse(parsedSql: object): DeparseReturnValues {
   try {
-    const data = deparse(parsedSql)
+    const data = deparse(parsedSql, {})
```
kevcodez, line 36: Types were erroring out, we can alternatively do ts-ignore

kamilogorek: Fixed it using class directly instead, as it's the same thing under the hood: https://github.com/launchql/pgsql-parser/blob/f1df82ed4358e47c682e007bc5aa306b58f25514/packages/deparser/src/deparser.ts#L258-L260

kamilogorek (approved):

kamilogorek:
Nothing that was used other than `customLogger` https://fastify.dev/docs/latest/Guides/Migration-Guide-V5/

PR #885, opened 2025-02-05
fix: Close both apps to allow for process restart

kamilogorek:
Currently when the main `app` is closed, or any node.js termination signal is sent (which is hijacked by `closeWithGrace` https://github.com/mcollina/close-with-grace?tab=readme-ov-file#api), an `adminApp` instance will keep it hostage and won't allow ec2 to restart everything correctly.

This can be tested by updating `/health` endpoint to call `app.close()` and observing that process is still running due to `adminApp` server being bound.

This will hopefully kill server instantly and won't let other requests come in to this instance as observed in our ec2 logs now.

Sidenote: why do we expose prometheus metrics on a separate port/app and not just `/metrics` endpoint within same app instead? @soedirgo

### DIFF
--- a/src/server/admin-app.ts
+++ b/src/server/admin-app.ts
@@ -1,11 +1,9 @@
 import { fastify, FastifyInstance, FastifyServerOptions } from 'fastify'
-
-import FastifyMetrics from 'fastify-metrics'
+import fastifyMetrics from 'fastify-metrics'
 
 export function build(opts: FastifyServerOptions = {}): FastifyInstance {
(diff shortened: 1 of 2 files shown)


pcnc (approved):

pcnc:
> Sidenote: why do we expose prometheus metrics on a separate port/app and not just /metrics endpoint within same app instead

I think it's so that we don't expose any type of metrics (not aware of the types of metrics we're exposing) - but eitherway those aren't being scraped either atm

PR #898, opened 2025-02-22
fix: proper nullable types for Go code gen

### LINKED ISSUE #897: Go Generated Nullable Types Are Not Serializable
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

These lines define the nullable types:
https://github.com/supabase/postgres-meta/blob/b85bf01fe51789588b87b4552d593b5510632bf9/src/server/templates/go.ts#L277C1-L288C2

The problem is, Go cannot automatically Decode into an object that uses the sql nullable variables. The modern approach is to use pointers, which can be nullable. 

## To Reproduce

In the below example, I have copied (a) the generated struct using supabase cli, (b) a sample response from a supabase rest call, and (c) the standard approach to decoding into a go object. As you can see, the error returns:

```
json: cannot unmarshal string into Go struct field PublicUsersSelect.date_of_birth of type sql.NullString
```

Run Online: https://go.dev/play/p/hOkZo4xt0bo

```go
package main

import (
	"database/sql"
	"encoding/json"
	"log"
	"strings"
)

// generated by supabase; copied here for demo
type PublicUsersSelect struct {
	ActiveOrganizationId sql.NullString `json:"active_organization_id"`
	Avatar               sql.NullString `json:"avatar"`
	DateOfBirth          sql.NullString `json:"date_of_birth"`
	FirstName            string         `json:"first_name"`
	FullName             string         `json:"full_name"`
	Id                   string         `json:"id"`
	IsSetup              bool           `json:"is_setup"`
	LastName             string         `json:"last_name"`
}

func main() {
	// usually from api; hard-coded here for demo
	res := `{"id":"35527ac4-dfd0-4743-89ab-a3ea226e4466","first_name":"Nicholas","last_name":"Barrow","full_name":"Nicholas Barrow","date_of_birth":"2002-02-11","avatar":null,"active_organization_id":"c7b65312-45b9-4076-a8c5-f9054fef435c","is_setup":true}`

	var user PublicUsersSelect
	err := json.NewDecoder(strings.NewReader(res)).Decode(&user)
	if err != nil {
		log.Println(err)
	}
}
```

## Expected behavior

The modern approach is to use pointers, which can be nullable. Another approach is to use custom wrappers with initializers, but this seems more cumbersome.

barrownicholas:

Bug Fix: resolves #897 

Nullable types are not de-serializable

## What is the new behavior?

Null-pointers allow proper serialization and better mimic proper use-case

## Additional context

See #897 for full breakdown


soedirgo (approved):
Thanks!

barrownicholas:
@soedirgo can you re-approve the tests for me? I had to re-factor a broken test.

barrownicholas:
@soedirgo looks like it wants another approval

soedirgo (approved):

Issue #831, opened 2024-11-07
Types generated by supabase db reset and supabase gen types are not consistent


saint-james-fr:
### Problem Summary

When running `supabase db reset`, I notice unexpected changes in the generated types file, even for columns that were not modified through migrations. This issue causes problems when comparing the generated types, as the differences seem random and inconsistent.

Specifically, the changes appear within a view that references the same table twice, in two Common Table Expressions (CTEs) or subqueries. The columns `id` from the table are ultimately combined into the view as `initial_contract_id` and `last_contract_id`.

### Scenario

Imagine a view that aggregates contract data. The view handles two distinct contract types: the **initial contract** and the **last contract**, both originating from the same `contract` table. These contracts are differentiated by their `id` and reunited based on the `contract_key`.

### Example of the Problem

The issue arises because the `supabase gen types` command generates different types on these columns even they're not expected - even when no direct changes have been made to the migrations or the underlying table structure. The position of the `referencedColumns` in the generated types file switches randomly between `initial_contract_id` and `last_contract_id`.

### Steps to Reproduce

1. Run `supabase db reset`.
2. Run the command to regenerate the types:
   ```bash
   supabase gen types --lang=typescript --local --schema public,not_public,storage > ../../packages/shared/src/types/supabase.gen.ts
   ```
3. Compare the generated types file (`supabase.gen.ts`), and you'll see the random changes in the order of `referencedColumns`.

#### Example Diff:

```diff
diff --git a/packages/shared/src/types/supabase.gen.ts b/packages/shared/src/types/supabase.gen.ts
index 40d7ea2..85ac078 100644
--- a/packages/shared/src/types/supabase.gen.ts
+++ b/packages/shared/src/types/supabase.gen.ts
@@ export type Database = {
             columns: ["contract_id"]
             isOneToOne: false
             referencedRelation: "view_contract"
-            referencedColumns: ["last_contract_id"]
+            referencedColumns: ["initial_contract_id"]
           },
           {
             foreignKeyName: "advance_payment_request_contract_id_fkey"
             columns: ["contract_id"]
             isOneToOne: false
             referencedRelation: "view_contract"
-            referencedColumns: ["initial_contract_id"]
+            referencedColumns: ["last_contract_id"]
           },
           {
             foreignKeyName: "advance_payment_request_contract_id_fkey"
@@ -2744,14 +2744,14 @@ export type Database = {
             columns: ["contract_id"]
             isOneToOne: false
[...]

PR #903, opened 2025-03-02
Fix: Deterministic  Referenced column sort to fix views referencing same table

saint-james-fr:
Bug fix: more deterministic sort to provide consistent output of typescript type generation

It's sorted by foreign_key_name, if they're the same then it's sorted by referenced_relation.

## Pain

Generating new types, unexpected changes occur in the generated types file, even for columns that were not modified through migrations. This issue causes problems when comparing the generated types, as the differences seem random and inconsistent. Specifically, the changes appear within a view that references the same table twice, in two Common Table Expressions (CTEs) or subqueries.

How it affected me? Comparing generated types in a CI (so types are always up to date to the latest migrations) was failing for random reasons as the generated typings were not consistent.

Issue [#831 on Supabase/CLI then transferred to supabase/postgres-meta](https://github.com/supabase/postgres-meta/issues/831) and [Issue supabase/postgres-meta#908 on supabase](https://github.com/supabase/postgres-meta/issues/908)

This PR addresses the problem by adding an additional sorting criterion to ensure a stable order of referencedColumns.

## What is the new behavior?

The new behavior ensures consistent and stable type generation. The sorting logic has been updated to ensure a deterministic order, preventing unintended changes in the generated types file.

It's sorted by foreign_key_name, if they're the same it's sorted by referenced_relation, then if they're the same it iterates over referenced_columns array to find the first difference and use it to sort and provide a deterministic order.

Finally if all columns are the same, it uses the length of the array to find a sorting order : i think this part could be improved and i'm open to any ways to improve it.

## Additional context

To verify the fix, I created a view that references the same table twice through Common Table Expressions (CTEs). I also updated the test cases to ensure that the initial_id block always appears below the second_id block, with initial_id being sorted before second_id.

Adding this view also led me to update relevant Swift and Go generated type code.


soedirgo (approved):
Thanks @saint-james-fr!

Issue #833, opened 2024-07-03
Improve generated TS schemas with `supabase gen types`


warrenbhw:
I'm using supabase auth, supabase RLS, and the supabase postgREST api. I like the idea of pushing authorization and validation rules all into the db, rather than having them split across the app layer and db.

However, one of the biggest points of friction that I encounter when building this way is reflecting the schema of my supabase tables and functions in my NextJS typescript codebase.

I use the supabase gen types command built into the Supabase CLI, but I end up having to write a lot of additional overrides, because not everything I want is inferred. Here are some examples:

- DB function return types are always unknown - I end up having to write them myself and override the generated type
- JSON types are just generic JSON, even when a better type could be inferred, ex. { id: string, name: string } - should be able to infer these from function definitions or from JSONschema check constraints
- Does not autogenerate schema definitions for zod/yup/valibot, so I end up having to write these manually any time I have a form
- For custom enum types, the type generation works well, but for domain types (aka refinement types), the type is inferred as unknown. Even if the generator doesn't want to generate zod/yup/valibot schemas for these, it could at least generate the base type of the domain instead of unknown.

Given these limitations, I feel that it would've been faster to just use Supabase as a vanilla Postgres and use Prisma - this way, I could define schemas once in typescript and get autogenerated zod, sql, and ts types.

Addressing these limitations would make me significantly more productive using Supabase w/ RLS+auth+postgREST.

Are there any good community projects that offer improved typegen, or are there plans to improve the type generator in supabase cli?
right now, it feels like i either have to get off the "all supabase" stack, or write my own type generator if I want to stay productive

PR #904, opened 2025-03-03
fix(template): dynamically set default schema alias in typescript type generation

bug-author:

This PR fixes the incorrect default alias generation for TypeScript type schemas when using a single non‑public schema with `supabase gen types`.

Currently, the TypeScript template always hardcodes the default schema alias as:

```ts
export type PublicSchema = Database[Extract<keyof Database, "public">]
```

This means that when a user generates types using a non‑public schema (for example, `--schema robo`), the generated helper types (for example. `Tables<"test_table">`) incorrectly reference the public schema, even if no tables exist there.

Relevant issues: 
- cli:  [#3146](https://github.com/supabase/cli/issues/3146), [#3061](https://github.com/supabase/cli/issues/3061)
- postgres-meta: #833 (for additional context)

## What is the new behavior?

This PR updates the TypeScript template in pg‑meta so that when exactly one non‑public schema is provided, the template automatically uses the environment variable `PG_META_GENERATE_TYPES_DEFAULT_SCHEMA` to generate a default alias. For example, if the schema provided is "robo", the template will now output:

```ts
export type DefaultSchema = Database[Extract<keyof Database, "robo">]
export type DefaultSchemaOrPublic = DefaultSchema
```

The helper types (`Tables`, `TablesInsert`, etc.) reference `DefaultSchemaOrPublic` instead of the hardcoded "public". In multi‑schema scenarios, the environment variable is not set, and the default behavior (using "public") remains, the users can explicitly specify the schema for non-public tables.

This change ensures:

- In single‑schema mode with a non‑public schema, shorthand type references (like `Tables<"test_table">`) correctly resolve to `Database["robo"]["Tables"]["test_table"]`.
- In multi‑schema mode, users must explicitly specify which schema to use (for example, `Tables<{ schema: "robo" }, "test_table">`).

## Additional context
Usage example

```
CREATE SCHEMA robo;
CREATE TABLE robo.test_table (
  id SERIAL PRIMARY KEY,
  name TEXT
);


CREATE TABLE public.example_table (
  id SERIAL PRIMARY KEY,
  description TEXT
);
```

```ts
// Scenario 1: No schema flag (defaults to "public")
import * as PublicTypes from "./dbTypes-public";

type PublicExampleRow = PublicTypes.Tables<"example_table">;

// Scenario 2: Single non‑public schema (e.g. "robo")
import * as RoboTypes from "./dbTypesRobo";

type RoboExampleRow = RoboTypes.Tables<"test_table">;

// Scenario 3: Multiple schemas specified (e.g. "public" and "robo")
import * as MultiTypes from "./dbTypes-multi";

type MultiPublicRow = MultiTypes.Tables<"example_table">;
type MultiRoboRow = MultiTypes.Tables<{ schema: "robo" }, "test_table">;

// examples
const publicRow: PublicExampleRow = {
  id: 1,
  description: "A public record",
};


const roboRow: RoboExampleRow = {
  id: 10,
  name: "Robo record",
};

const multiPublicRow: MultiPublicRow = {
  id: 2,
  description: "Multi public record",
};

const multiRoboRow: MultiRoboRow = {
  id: 20,
  name: null,
};
```

Corresponding [cli PR](https://github.com/supabase/cli/pull/3243)

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -8,8 +8,7 @@ import type {
   PostgresView,
 } from '../../lib/index.js'
 import type { GeneratorMetadata } from '../../lib/generators.js'
-
-const defaultSchema = process.env.PG_META_GENERATE_TYPES_DEFAULT_SCHEMA
+import { DEFAULT_SCHEMA } from '../constants.js'
 
 export const apply = async ({
   schemas,
@@ -392,18 +391,12 @@ export type Database = {
     })}
 }
 
-${
-  defaultSchema
-    ? `export type DefaultSchema = Database[Extract<keyof Database, ${JSON.stringify(defaultSchema)}>]
-  export type DefaultSchemaOrPublic = DefaultSchema`
-    : `export type PublicSchema = Database[Extract<keyof Database, "public">]
-  export type DefaultSchemaOrPublic = PublicSchema`
-}
+ type DefaultSchema = Database[Extract<keyof Database, ${JSON.stringify(DEFAULT_SCHEMA) || '"public"'}>]
 
 
 export type Tables<
   PublicTableNameOrOptions extends
-    | keyof (DefaultSchemaOrPublic["Tables"] & DefaultSchemaOrPublic["Views"])
+    | keyof (DefaultSchema["Tables"] & DefaultSchema["Views"])
     | { schema: keyof Database },
   TableName extends PublicTableNameOrOptions extends { schema: keyof Database }
     ? keyof (Database[PublicTableNameOrOptions["schema"]]["Tables"] &
@@ -416,8 +409,8 @@ export type Tables<
     }
     ? R
     : never
-  : PublicTableNameOrOptions extends keyof (DefaultSchemaOrPublic["Tables"] & DefaultSchemaOrPublic["Views"])
-    ? (DefaultSchemaOrPublic["Tables"] & DefaultSchemaOrPublic["Views"])[PublicTableNameOrOptions] extends {
+  : PublicTableNameOrOptions extends keyof (DefaultSchema["Tables"] & DefaultSchema["Views"])
+    ? (DefaultSchema["Tables"] & DefaultSchema["Views"])[PublicTableNameOrOptions] extends {
         Row: infer R
       }
       ? R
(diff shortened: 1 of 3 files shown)


src/server/templates/typescript.ts
```diff
@@ -9,6 +9,8 @@ import type {
 } from '../../lib/index.js'
 import type { GeneratorMetadata } from '../../lib/generators.js'
 
+const defaultSchema = process.env.PG_META_GENERATE_TYPES_DEFAULT_SCHEMA
```
soedirgo, line 12: This should be in `src/server/constants.ts`

src/server/templates/typescript.ts
```diff
@@ -390,11 +392,18 @@ export type Database = {
     })}
 }
 
-type PublicSchema = Database[Extract<keyof Database, "public">]
+${
+  defaultSchema
+    ? `export type DefaultSchema = Database[Extract<keyof Database, ${JSON.stringify(defaultSchema)}>]
+  export type DefaultSchemaOrPublic = DefaultSchema`
+    : `export type PublicSchema = Database[Extract<keyof Database, "public">]
+  export type DefaultSchemaOrPublic = PublicSchema`
```
soedirgo, line 400: This should be kept private and unexported

src/server/templates/typescript.ts
```diff
@@ -390,11 +392,18 @@ export type Database = {
...
+}
+
 
 export type Tables<
   PublicTableNameOrOptions extends
-    | keyof (PublicSchema["Tables"] & PublicSchema["Views"])
+    | keyof (DefaultSchemaOrPublic["Tables"] & DefaultSchemaOrPublic["Views"])
```
soedirgo, line 406: Better to use `DefaultSchema` nomenclature everywhere vs. `PublicSchema`

soedirgo:
Thanks @bug-author! Left some comments

src/server/templates/typescript.ts
bug-author: Fixed

src/server/templates/typescript.ts
bug-author: This is now refactored to 
```ts
type DefaultSchema = Database[Extract<keyof Database, ${JSON.stringify(DEFAULT_SCHEMA) || '"public"'}>]
```

src/server/templates/typescript.ts
bug-author: Resolved

bug-author:
@soedirgo Thank you for the review. I have made the suggested changes:

- Moved the default schema constant to `src/server/constants.ts`
- Refactored schema aliases to use `DefaultSchema` only

Thanks for helping me refine the implementation.

soedirgo (approved):
Thanks again @bug-author!

PR #911, opened 2025-03-25
feat(server): introduce maxResultSize limit and fix pg errors handling

avallete:

- Fixes the way we handled pg.Pool non "database" error (connection level errors) to avoid throwing unhandled exception, but instead make sure it get caught and propagated to the caller in the response via `poolerQueryHandleError`
- Going along with https://github.com/supabase/node-postgres/pull/1 introducing a way to cap the maximum size we allow a query to take in memory before aborting with an error.

With those changes the running server should be more robust, die less often and leverage killing the connections sooner than later.

Fixes PGMETA-45


src/lib/db.ts
```diff
@@ -21,6 +20,38 @@ pg.types.setTypeParser(1185, parseArray) // _timestamptz
...
+    const connectionErrorHandler = (err: any) => {
+      // If the error hasn't already be propagated to the catch
+      if (!rejected) {
+        rejected = true
+        reject(err)
+      }
+    }
+    // This listened avoid getting uncaught exceptions for errors happening at connection level within the stream
+    // such as parse or RESULT_SIZE_EXCEEDED errors instead, handle the error gracefully by bubbling in up to the caller
+    pool.once('error', connectionErrorHandler)
```
avallete, line 37: **note**

Without this, by default, `node-postgres` will propagate the error within the catch, but ALSO in the global scope. Which, when within a server, will make the whole server die.

This wrapper, and listener make sure that every error that is not handled within the catch is propagated as a promise rejection. And the errors that have already been handled by the promises are dismissed and doesn't bubble up the stack.

src/lib/db.ts
```diff
@@ -60,26 +91,27 @@ export const init: (config: PoolConfig) => {
...
 
-        let res = await pool.query(sql)
+        let res = await poolerQueryHandleError(pool, sql)
         if (Array.isArray(res)) {
           res = res.reverse().find((x) => x.rows.length !== 0) ?? { rows: [] }
         }
         return { data: res.rows, error: null }
       } catch (error: any) {
-        if (error instanceof DatabaseError) {
+        if (error.constructor.name === 'DatabaseError') {
```
avallete, line 114: **note**

With the new error propagation, had to slightly tweak this as the `instanceof` wasn't properly working. Even tough the constructor was the right one, the instance kept being an `Error` because of the promise wrapping.

src/lib/db.ts
```diff
@@ -147,7 +179,60 @@ ${' '.repeat(5 + lineNumber.toString().length + 2 + lineOffset)}^
...
+          const _pool = pool
+          pool = null
+          try {
+            if (_pool) {
+              // Force kill the connection by destroying the socket
+              const client = (_pool as any)._clients?.[0]
+              if (client?.connection?.stream) {
+                client.connection.stream.destroy()
+              }
+            }
```
avallete, line 194: **note**

This is a bit dirty, but is the quickest way to kill the connection, making sure the GC can go and collect the "results" data to free up the memory.

avallete (commented):
This will need more testing on staging, and will need close monitoring on prod deploy. But should help reduce the spikes of 5xx errors we see.

soedirgo (approved):

PR #914, opened 2025-03-30
fix(typescript): function only have union over args

avallete:

- A function definition in postgresql can be declared multiples times with differents arguments. But it's return type must be the same across all definition. Therefore, the generated types should match that. This turn multiples definitions like so:

```ts
get_todos_setof_rows: { Args: { ... }, Returns: { ... }  } | { Args: { ... }, Returns: { ... }  }
```
Into the more correct:

```ts
get_todos_setof_rows: { Args: { ... } | { ... }, Returns: { ... }  }
```

This also will fix on the type inference side the return type for `rpc` functions calls from an union of object, to a single object.

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -277,88 +277,74 @@ export type Database = {
 
               return Object.entries(schemaFunctionsGroupedByName).map(
                 ([fnName, fns]) =>
-                  `${JSON.stringify(fnName)}: ${fns
-                    .map(
-                      ({
-                        args,
-                        return_type_id,
-                        return_type_relation_id,
-                        is_set_returning_function,
-                      }) => `{
-                  Args: ${(() => {
-                    const inArgs = args.filter(({ mode }) => mode === 'in')
+                  `${JSON.stringify(fnName)}: {
+                      Args: ${fns
+                        .map(({ args }) => {
+                          const inArgs = args.filter(({ mode }) => mode === 'in')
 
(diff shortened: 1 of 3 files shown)


src/server/templates/typescript.ts
```diff
@@ -277,88 +277,74 @@ export type Database = {
...
-                      const type = types.find(({ id }) => id === type_id)
-                      let tsType = 'unknown'
-                      if (type) {
-                        tsType = pgTypeToTsType(type.name, { types, schemas, tables, views })
-                      }
-                      return { name, type: tsType, has_default }
-                    })
+                  `${JSON.stringify(fnName)}: {
+                      Args: ${fns
+                        .map(({ args }) => {
```
avallete, line 282: **note**

The loop is now at the `Args` level to unionize.

src/server/templates/typescript.ts
```diff
@@ -277,88 +277,74 @@ export type Database = {
...
+                              tsType = pgTypeToTsType(type.name, { types, schemas, tables, views })
+                            }
+                            return { name, type: tsType }
+                          })
+
+                          return `{
+                            ${argsNameAndType.map(
+                              ({ name, type }) => `${JSON.stringify(name)}: ${type}`
+                            )}
+                          }`
```
avallete, line 318: **note**

Single `Returns` using the first element within the `fns` array as source of truth.

soedirgo (approved):

PR #920, opened 2025-04-02
Add tests for less common entity routes

olirice:

WARNING! LLM Generated, please review carefully

Test coverage for
server/config
server/extensions
server/format
server/functions
server/generators
server/policies
server/publications
server/triggers
server/types

to get coverage > 80%

### DIFF
--- a/src/lib/PostgresMetaPolicies.ts
+++ b/src/lib/PostgresMetaPolicies.ts
@@ -112,6 +112,9 @@ export default class PostgresMetaPolicies {
     command?: string
     roles?: string[]
   }): Promise<PostgresMetaResult<PostgresPolicy>> {
+    if (!table || !name) {
+      return { data: null, error: { message: 'Missing required name or table parameter' } }
+    }
     const definitionClause = definition === undefined ? '' : `USING (${definition})`
     const checkClause = check === undefined ? '' : `WITH CHECK (${check})`
     const sql = `
--- a/test/server/format.ts
+++ b/test/server/format.ts
@@ -10,9 +10,16 @@ test('format SQL query', async () => {
   expect(res.statusCode).toBe(200)
   expect(res.headers['content-type']).toContain('text/plain')
   const formattedQuery = res.body
-  expect(formattedQuery).toContain('SELECT')
-  expect(formattedQuery).toContain('FROM')
-  expect(formattedQuery).toContain('WHERE')
+  expect(formattedQuery).toMatchInlineSnapshot(`
+    "SELECT
+      id,
+      name
+    FROM
+      users
+    WHERE
+      status = 'ACTIVE'
+    "
+  `)
 })
 
 test('format complex SQL query', async () => {
@@ -26,7 +33,24 @@ test('format complex SQL query', async () => {
   })
   expect(res.statusCode).toBe(200)
   expect(res.headers['content-type']).toContain('text/plain')
-  expect(res.body).toBeTruthy()
+  expect(res.body).toMatchInlineSnapshot(`
+    "SELECT
+      u.id,
+      u.name,
+      p.title,
+      p.created_at
+    FROM
+      users u
+      JOIN posts p ON u.id = p.user_id
+    WHERE
+      u.status = 'ACTIVE'
+      AND p.published = true
+    ORDER BY
+      p.created_at DESC
+    LIMIT
+      10
+    "
+  `)
 })
 
 test('format invalid SQL query', async () => {
@@ -35,12 +59,17 @@ test('format invalid SQL query', async () => {
     path: '/query/format',
     payload: { query: 'SELECT FROM WHERE;' },
   })
-  // Even invalid SQL can be formatted
   expect(res.statusCode).toBe(200)
   expect(res.headers['content-type']).toContain('text/plain')
-  expect(res.body).toBeTruthy()
+  expect(res.body).toMatchInlineSnapshot(`
+    "SELECT
+    FROM
+    WHERE;
+    "
+  `)
 })
 
+// TODO(andrew): Those should return 400 error code for invalid parameter
 test('format empty query', async () => {
   const res = await app.inject({
     method: 'POST',
--- a/test/server/generators.ts
+++ b/test/server/generators.ts
@@ -8,7 +8,7 @@ test('typescript generator route', async () => {
   })
   expect(res.statusCode).toBe(200)
   expect(res.headers['content-type']).toContain('text/plain')
-  expect(res.body).toBeTruthy()
+  expect(res.body).contain('public')
 })
 
 test('go generator route', async () => {
@@ -34,22 +34,18 @@ test('swift generator route', async () => {
 test('generator routes with includedSchemas parameter', async () => {
   const res = await app.inject({
     method: 'GET',
-    path: '/generators/typescript?includedSchemas=public',
+    path: '/generators/typescript?included_schemas=private',
   })
   expect(res.statusCode).toBe(200)
   expect(res.headers['content-type']).toContain('text/plain')
-  expect(res.body).toBeTruthy()
+  // the only schema is excluded database should be empty
+  expect(res.body).toContain('Database = {}')
 })
 
-// Skip this test as the OpenAPI endpoint is not implemented
-test.skip('openapi generator route', async () => {
+test('invalid generator route', async () => {
   const res = await app.inject({
     method: 'GET',
     path: '/generators/openapi',
   })
-  expect(res.statusCode).toBe(200)
-  expect(res.headers['content-type']).toContain('application/json')
-  const body = JSON.parse(res.body)
-  expect(body.openapi).toBeTruthy()
-  expect(body.paths).toBeTruthy()
+  expect(res.statusCode).toBe(404)
 })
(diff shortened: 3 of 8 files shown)


src/lib/PostgresMetaPolicies.ts
```diff
@@ -112,6 +112,9 @@ export default class PostgresMetaPolicies {
     command?: string
     roles?: string[]
   }): Promise<PostgresMetaResult<PostgresPolicy>> {
+    if (!table || !name) {
+      return { data: null, error: { message: 'Missing required name or table parameter' } }
+    }
```
avallete, line 117: **chore**

Avoid `ident` to raise an exception ending with `500` if those params are empty. TODO: migrate all this to zod for better validation of the params.

avallete (approved):

test/server/generators.ts
```diff
@@ -0,0 +1,51 @@
+import { expect, test } from 'vitest'
+import { app } from './utils'
+
+test('typescript generator route', async () => {
```
egor-romanov, line 4: not actually testing anything

test/server/generators.ts
```diff
@@ -0,0 +1,51 @@
...
+  const res = await app.inject({
+    method: 'GET',
+    path: '/generators/typescript',
+  })
+  expect(res.statusCode).toBe(200)
+  expect(res.headers['content-type']).toContain('text/plain')
+  expect(res.body).contain('public')
+})
+
+test('go generator route', async () => {
```
egor-romanov, line 14: same, not a test

test/server/generators.ts
```diff
@@ -0,0 +1,51 @@
...
+  const res = await app.inject({
+    method: 'GET',
+    path: '/generators/go',
+  })
+  expect(res.statusCode).toBe(200)
+  expect(res.headers['content-type']).toContain('text/plain')
+  expect(res.body).toBeTruthy()
+})
+
+test('swift generator route', async () => {
```
egor-romanov, line 24: same, misses essential checks

test/server/generators.ts
```diff
@@ -0,0 +1,51 @@
...
+  const res = await app.inject({
+    method: 'GET',
+    path: '/generators/swift',
+  })
+  expect(res.statusCode).toBe(200)
+  expect(res.headers['content-type']).toContain('text/plain')
+  expect(res.body).toBeTruthy()
+})
+
+test('generator routes with includedSchemas parameter', async () => {
```
egor-romanov, line 34: valid, but a lot of actual issues in the code may be missed, would be better to include one or a few of multiple existing schemas and check that only they are included

test/server/policies.ts
```diff
@@ -0,0 +1,71 @@
+import { expect, test } from 'vitest'
+import { app } from './utils'
+
+test('policy list filtering', async () => {
```
egor-romanov, line 4:
```suggestion
test('policy list with limit', async () => {
```

egor-romanov: also same as for other list tests, we should create some first to check that they are returned and that limit works

test/server/policies.ts
```diff
@@ -0,0 +1,71 @@
...
+    method: 'GET',
+    path: '/policies?limit=5',
+  })
+  expect(res.statusCode).toBe(200)
+  const policies = res.json()
+  expect(Array.isArray(policies)).toBe(true)
+  expect(policies.length).toBeLessThanOrEqual(5)
+})
+
+test('policy list with specific included schema', async () => {
```
egor-romanov, line 15: we need to create some in public and in other schema(s) first (we prob have only 1)

test/server/triggers.ts
```diff
@@ -0,0 +1,75 @@
+import { expect, test } from 'vitest'
+import { app } from './utils'
+
+test('trigger list filtering', async () => {
```
egor-romanov, line 4:
```suggestion
test('trigger list with limit', async () => {
```

egor-romanov: also same as for other list tests, we should create some first to check that they are returned and that limit works

test/server/triggers.ts
```diff
@@ -0,0 +1,75 @@
...
+    method: 'GET',
+    path: '/triggers?limit=5',
+  })
+  expect(res.statusCode).toBe(200)
+  const triggers = res.json()
+  expect(Array.isArray(triggers)).toBe(true)
+  expect(triggers.length).toBeLessThanOrEqual(5)
+})
+
+test('trigger list with specific included schema', async () => {
```
egor-romanov, line 15: we need to create some in public and in other schema(s) first (maybe we have one from migrations in public)

test/server/triggers.ts
```diff
@@ -0,0 +1,75 @@
...
+  expect(res.statusCode).toBe(200)
+  const triggers = res.json()
+  expect(Array.isArray(triggers)).toBe(true)
+  // All triggers should be in the public schema
+  triggers.forEach((trigger) => {
+    expect(trigger.schema).toBe('public')
+  })
+})
+
+test('trigger with invalid id', async () => {
```
egor-romanov, line 29:
```suggestion
test('get trigger with invalid id', async () => {
```

test/server/types.ts
```diff
@@ -0,0 +1,62 @@
...
+    method: 'GET',
+    path: '/types?limit=5',
+  })
+  expect(res.statusCode).toBe(200)
+  const types = res.json()
+  expect(Array.isArray(types)).toBe(true)
+  expect(types.length).toBeLessThanOrEqual(5)
+})
+
+test('type list with specific included schema', async () => {
```
egor-romanov, line 15: we need to create some in other schema(s) first

test/server/types.ts
```diff
@@ -0,0 +1,62 @@
+import { expect, test } from 'vitest'
+import { app } from './utils'
+
+test('type list filtering', async () => {
```
egor-romanov, line 4: also same as for other list tests, we should create some first to check that they are returned and that limit works (not sure we have >5)

egor-romanov, line 4:
```suggestion
test('type list with limit', async () => {
```

test/server/types.ts
```diff
@@ -0,0 +1,62 @@
...
+test('type list excluding array types', async () => {
+  const res = await app.inject({
+    method: 'GET',
+    path: '/types?includeArrayTypes=false',
+  })
+  expect(res.statusCode).toBe(200)
+  const types = res.json()
+  expect(Array.isArray(types)).toBe(true)
+  // Should not include array types
+  const arrayTypes = types.filter((type) => type.name.startsWith('_'))
```
egor-romanov, line 38: do we have array types? 
sorry if missed it

test/server/format.ts
```diff
@@ -0,0 +1,89 @@
...
+  expect(res.headers['content-type']).toContain('text/plain')
+  expect(res.body).toMatchInlineSnapshot(`
+    "SELECT
+    FROM
+    WHERE;
+    "
+  `)
+})
+
+// TODO(andrew): Those should return 400 error code for invalid parameter
```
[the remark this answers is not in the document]
soedirgo: Yeah agreed 200 is fine here. I think just like formatters in code editors it shouldn't fail when there are syntax errors; rather it should try to format the rest of the code despite the errors.

test/server/generators.ts
avallete: I would actually disagree, even if minimal this ensure that the route doesn't throw an exception (which occurs if invalid TS is generated when passing within prettier for instance, or if there is a bug in the JS within the string interpolations used to generate the content).

That's not a great test, but that's a start.

egor-romanov: and if due to code changes it this route starts to return no types instead of an error, can you spot it with this test?

this test can be sort of easily improved if you have a file with types saved somewhere and you compare the body with the template file (it can even be contains and some expected parts)

test/server/functions.ts
```diff
@@ -0,0 +1,81 @@
+import { expect, test } from 'vitest'
+import { app } from './utils'
+
+test('function list filtering', async () => {
```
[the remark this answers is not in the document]
egor-romanov: yeah, that is not great, but it is not the only way, another one is “create functions - test - drop functions”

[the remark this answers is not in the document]
avallete: Issue with that is that still cause conflicts depending on the orders the tests run in (sometimes the drop didn't happen while another introspection query will get a different result). I guess we can make the tests more synchronous to avoid that, even if I'm not sure how, maybe wrap them in `describe` ?

test/server/generators.ts
avallete: Yeah, that's not great. I guess in the case you mention "public" will probably not be present in the "schemas" hence fail the test. But I agree that's brittle.

Looking at your suggestion this might be a good place to use `toMatchInlineSnapshot` so this can still be updated easily if types generation changes and make it more easily spotable in reviews.

test/server/functions.ts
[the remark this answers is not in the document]
egor-romanov: i mean, it prob wont conflict if you still check that there are exactly 5 functions and create 6 in advance, you dont have to check function names

for a better test we can also query functions from the database directly and check that pg-meta returns a subset of the functions that we got using db query directly

test/server/generators.ts
egor-romanov: we can just create a table, a function, a view, etc in the beginning of the test and check that types are generated as expected for these

PR #926, opened 2025-04-09
fix: add more logs to the exit path

avallete:
So we can understand better what cause the ending of the container

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -154,28 +154,58 @@ if (EXPORT_DOCS) {
 } else if (GENERATE_TYPES) {
   console.log(await getTypeOutput())
 } else {
-  const closeListeners = closeWithGrace(async ({ err }) => {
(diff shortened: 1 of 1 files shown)


src/server/server.ts
```diff
@@ -154,28 +154,58 @@ if (EXPORT_DOCS) {
...
-      app.log.error(err)
+      app.log.error({ err }, 'Uncaught error in app, exit(1)')
       process.exit(1)
     }
     const adminPort = PG_META_PORT + 1
-    adminApp.listen({ port: adminPort, host: PG_META_HOST })
+    adminApp.listen({ port: adminPort, host: PG_META_HOST }, (err) => {
+      if (err) {
+        app.log.error({ err }, 'Uncaught error in adminApp, exit(1)')
+        process.exit(1)
```
avallete, line 207: **note**


This one I'm not sure, should we exit 1 in such case ? It seems like this would happen anyways since we'll pass by the hook 'onClose' 🤔

PR #930, opened 2025-04-09
fix(typegen): use javascript deterministic sort for args join

avallete:

Should resolve: https://github.com/supabase/postgres-meta/issues/929


src/server/templates/typescript.ts
```diff
@@ -296,6 +296,7 @@ export type Database = {
                           })
                           return `{ ${argsNameAndType.map(({ name, type, has_default }) => `${JSON.stringify(name)}${has_default ? '?' : ''}: ${type}`)} }`
                         })
+                        .toSorted()
```
soedirgo, line 299: TIL `.toSorted()` 🔥

soedirgo (approved):

PR #932, opened 2025-04-11
fix: catch invalid url early

avallete:

- Catch if there is an invalid pg url early on in the request processing and unify the error message

Related: PGMETA-53

### DIFF
--- a/src/server/routes/index.ts
+++ b/src/server/routes/index.ts
@@ -55,7 +55,7 @@ export default async (fastify: FastifyInstance) => {
         request.log.error({ message: 'pg connection string is invalid url' })
         throw new Error('failed to process upstream connection details')
       }
-      done()
+      return done()
     } catch (err) {
       return done(err as Error)
     }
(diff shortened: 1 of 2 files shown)


src/server/routes/index.ts
```diff
@@ -26,24 +26,39 @@ import { PG_CONNECTION, CRYPTO_KEY } from '../constants.js'
...
+        } catch (e: any) {
+          request.log.warn({
+            message: 'failed to parse encrypted connstring',
+            error: e.toString(),
+          })
+          throw new Error('failed to process upstream connection details')
+        }
+      } else {
+        request.headers.pg = PG_CONNECTION
+      }
```
avallete, line 46: **note**

To identify why and how this happen, would it be okay to console.error the `x-connection-encrypted` ?

soedirgo: I think no, you can use the encrypted connection string to query projects you don't have access to on the dashboard

soedirgo: I think no, it's still sensitive information

soedirgo: I think no, safer not to log any potentially sensitive information

soedirgo (approved):

PR #934, opened 2025-04-14
fix(monitoring): add sentry monitoring

avallete:
For better performances tracing and error alerting use sentry

Closes: PGMETA-54

### DIFF
--- a/src/server/routes/query.ts
+++ b/src/server/routes/query.ts
@@ -21,7 +21,7 @@ export default async (fastify: FastifyInstance) => {
     const connectionString = request.headers.pg
 
     const pgMeta = new PostgresMeta({ ...DEFAULT_POOL_CONFIG, connectionString })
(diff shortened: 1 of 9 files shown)


src/server/routes/query.ts
```diff
@@ -21,7 +21,7 @@ export default async (fastify: FastifyInstance) => {
     const connectionString = request.headers.pg
 
     const pgMeta = new PostgresMeta({ ...DEFAULT_POOL_CONFIG, connectionString })
-    const { data, error } = await pgMeta.query(request.body.query)
+    const { data, error } = await pgMeta.query(request.body.query, true)
```
avallete, line 24: **note**

This'll ensure that custom queries from the SQL Editor won't bubble up. But will allow to provide details about the rest of the generated introspections queries. The rational behind that is that custom queries from the user might contain som PII that we don't want around in the tracing.

soedirgo (approved):

PR #947, opened 2025-05-21
fix: exclude pooler from statement_timeout

avallete:

## What is the new behavior?

## Additional context


src/lib/secrets.ts
```diff
@@ -15,6 +12,8 @@ export const getSecret = async (key: string) => {
   if (!file) {
     return ''
   }
+  // Use dynamic import to support module mock
+  const fs = await import('node:fs/promises')
```
avallete, line 16: **note**

The dynamic module import need to be within the function to avoid other imports of `constant.ts` to possibly break the mocks in the tests.

Performance impact should be minimal as we just use it at server startup time.

PR #958, opened 2025-07-22
fix(server): bump bodyLimit to 3MB default allow parameterize

avallete:

- Introduce `PG_META_MAX_BODY_LIMIT_MB` env variable to make easier parameterization
- Bump the default `bodyLimit` value from 1MB to 3MB.


soedirgo:
What’s the context around parameterization?

soedirgo (approved):

Issue #959, opened 2025-06-09
Generated types not consistently reproducable


dangbert:
**Describe the bug**
Since upgrading to Supbase CLI version 2.24.3 and upgrading my Supbase major_version from 15 -> 17 at the same time, my generated typescript types are flaky (the output file flips between one of two versions between successive generations).

**To Reproduce**

1. Init a supabase project locally with major_version 17 in config.toml and these npm package version:
````bash
├── eslint@9.28.0
├── supabase@2.24.3
├── @supabase/supabase-js@2.49.4
├── typescript-eslint@8.33.1
├── typescript@5.8.3
````

2. Run `npx supabase start` and create any empty table at http://localhost:54323/project/default/editor then run `npx supabase db dump --data-only --linked  -f ./supabase/seed.sql && npx supabase db diff -f "init"`

3. `npx supbase db reset && npx supabase gen types typescript --local > supabase.ts`

4. Rerun the previous command several times and observe that this section of `supbase.ts` varies between the following values:

Variation A:
````ts
export type Database = {
  graphql_public: {
    Tables: {
      [_ in never]: never
    }
    Views: {
      [_ in never]: never
    }
    Functions: {
      graphql: {
        Args: {
          operationName?: string
          query?: string
          variables?: Json
          extensions?: Json
        }
        Returns: Json
      }
    }
    Enums: {
      [_ in never]: never
    }
    CompositeTypes: {
      [_ in never]: never
    }
  }
...
````

and variation B:
````ts
export type Database = {
  graphql_public: {
    Tables: {
      [_ in never]: never
    }
    Views: {
      [_ in never]: never
    }
    Functions: {
      graphql: {
        Args: {
          variables?: Json
          extensions?: Json
          operationName?: string
          query?: string
        }
        Returns: Json
      }
    }
    Enums: {
      [_ in never]: never
    }
    CompositeTypes: {
      [_ in never]: never
    }
  }
...
````

Where the diff looks like this:

````bash
+++ b/supabase.ts
@@ -17,10 +17,10 @@ export type Database = {
     Functions: {
       graphql: {
         Args: {
-          operationName?: string
-          query?: string
           variables?: Json
           extensions?: Json
+          operationName?: string
+          query?: string
         }
         Returns: Json
       }
````

**Expected behavior**
The generated file, supabase.ts should be reproducible and exactly identical every time it's created (provided there are no changes to the database schema).

[...]

PR #964, opened 2025-07-30
fix(typegen): improve ts typegen consistency between generation

avallete:

- Improve typescript function args generation so it's more stable for the same function declaration
- Add a test to avoid regression, ensuring that the type generation is stable regardless of:
1. Entity order creation (eg: `CREATE TABLE A; CREATE TABLE B` or `CREATE TABLE B; CREATE TABLE A`) should always result in the same final types.
2. Entities properties order declaration (except for enums). Which wasn't the case for function args declaration (eg: `CREATE TABLE A (col1, col2)` or `CREATE TABLE A (col2, col1)`) should always result in the same final types.

Fixes: #959

### DIFF
--- a/test/server/typegen.ts
+++ b/test/server/typegen.ts
@@ -629,659 +629,6 @@ test('typegen: typescript', async () => {
   )
 })
 
-test('typegen w/ one-to-one relationships', async () => {
-  const { body } = await app.inject({
-    method: 'GET',
-    path: '/generators/typescript',
-    query: { detect_one_to_one_relationships: 'true' },
-  })
-  expect(body).toMatchInlineSnapshot(
-    `
-    "export type Json =
-      | string
-      | number
-      | boolean
-      | null
-      | { [key: string]: Json | undefined }
-      | Json[]
-
-    export type Database = {
-      public: {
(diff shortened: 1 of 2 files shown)


test/server/typegen.ts
```diff
@@ -629,659 +629,6 @@ test('typegen: typescript', async () => {
   )
 })
 
-test('typegen w/ one-to-one relationships', async () => {
```
avallete, line 632: This test was a duplicate with: https://github.com/supabase/postgres-meta/pull/964/files#diff-a9ec34b3928a1f6c1c476fc768d833dbe06c573bc6b9282719574f04f0b4d523L1285

test/server/typegen.ts
```diff
@@ -2593,6 +1940,273 @@ test('typegen: typescript w/ postgrestVersion', async () => {
   )
 })
 
+test('typegen: typescript consistent types definitions orders', async () => {
```
avallete, line 1943: In there we now have something that should warn us if a regression in generation stability occurs on any of the main entities.

avallete (commented):
Some comments for reviewers.

PR #971, opened 2025-08-05
feat(typegen): add functions setof type introspection

avallete:

## What is the new behavior?

## Additional context


src/lib/types.ts
```diff
@@ -156,6 +157,9 @@ const postgresFunctionSchema = Type.Object({
   return_type: Type.String(),
   return_type_relation_id: Type.Union([Type.Integer(), Type.Null()]),
   is_set_returning_function: Type.Boolean(),
+  returns_set_of_table: Type.Boolean(),
+  return_table_name: Type.Union([Type.String(), Type.Null()]),
+  returns_multiple_rows: Type.Boolean(),
```
soedirgo, line 162: Shouldn't these be covered by `return_type_relation_id` and `is_set_returning_function`?

src/lib/sql/functions.sql.ts
```diff
@@ -85,6 +85,25 @@ select
...
+  case
+    when rt.typrelid != 0 then
+      (select relname from pg_class where oid = rt.typrelid)
+    else null
+  end as return_table_name,
+  case
+    when f.proretset then
+      coalesce(f.prorows, 0) > 1
+    else false
+  end as returns_multiple_rows,
```
soedirgo, line 106: I'm thinking instead of `returns_multiple_rows` we can just return `prorows` as is (setting it as `null` if 0)

It's easy to confuse it with `is_set_returning_function` since they [mean the same thing](https://www.postgresql.org/docs/current/functions-srf.html#FUNCTIONS-SRF); rather what we're doing is adhering to PostgREST's idiosyncrasies

soedirgo:
Can we split out the refactor into another PR? It's difficult to see which change is related to computed relationships.

PR #973, opened 2025-08-08
Allow strings for `numeric` type inserts and updates

### LINKED ISSUE #972: For numeric columns, TypeScript type generation does not include `string` as an option for inserts and updates
# Bug report

## Describe the bug

When generating TypeScript types, the current code “correctly” converts `numeric` types to `number`, as Postgres and Postgrest return such columns as JSON numbers by default without a query cast to `::text`.

However, generated insert and update types are *limited* to `number`, whereas the underlying APIs will accept a `string` and preserve precision this way. As such, the generated update and insert types for these fields should allow for `string` parameters.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Define a table with column type `NUMERIC` or `DECIMAL`.
2. Generate TypeScript types for the table.
3. Observe that the `Row` type is `number`, but so are the `Insert` and `Update` types.

## Expected behavior

`number | string` is used instead for `Insert` and `Update` types.

## Additional context

Currently we're having to cast strings `as unknown as number` or `@ts-expect-error` for all insert/update calls to get around this.

Shadowfiend:
When generating TypeScript types, the current code "correctly" reports numeric types as `number`, as Postgres and Postgrest return such columns as JSON numbers by default without a query `::text` cast.

However, generated insert and update types are limited to `number`, whereas the underlying APIs will accept a string and allow preserving precision that would be lost in conversion to JS `number` this way.

Generated types for inserting and updating numeric fields now include `| string` as a possibility to allow for this.

Closes #972.

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -144,6 +144,16 @@ export type Database = {
                         views,
                       })
 
+                      if (column.name.toLowerCase().startsWith("decimal")) {
+                        console.log("YOYOYOYO Insert", `'${column.name}'`, `'${column.format}'`)
+                      }
+
+                      // Numeric types can be inserted as strings to preserve
+                      // precision.
+                      if (column.format === "numeric") {
+                        output += '| string'
+                      }
+
                       if (column.is_nullable) {
                         output += '| null'
                       }
@@ -165,6 +175,12 @@ export type Database = {
                         tables,
                         views,
                       })}`
+
+                      // Numeric types can be inserted as strings to preserve
+                      // precision.
+                      if (column.format === "numeric") {
+                        output += '| string'
+                      }
 
                       if (column.is_nullable) {
                         output += '| null'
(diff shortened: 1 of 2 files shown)


src/server/templates/typescript.ts
```diff
@@ -144,6 +144,16 @@ export type Database = {
                         views,
                       })
 
+                      if (column.name.toLowerCase().startsWith("decimal")) {
+                        console.log("YOYOYOYO Insert", `'${column.name}'`, `'${column.format}'`)
+                      }
```
avallete, line 149: **issue**

Probably don't want this to stay around.

avallete (changes requested):
Thanks for the contribution! I appreciate the effort you put into this PR.

That said, I’m not in favor of merging this as a global change. In most cases, this approach could introduce issues down the line, especially since it primarily affects bigint and similar fields. And could also constitute a breaking change in user code where functions might expect things to be a `number` and not a `number | string` type. Instead, I’d recommend handling this on a per-column basis [like mentioned in the guide](https://supabase.com/docs/guides/api/rest/generating-types#helper-types-for-tables-and-joins), if you know a specific column may have precision issues when converted in JavaScript, you can override it to be typed as `number | string` or `number | bigint`.

This way, we avoid broad changes that might cause unintended side effects while still giving flexibility where needed.

avallete:

PR #980, opened 2025-08-29
perf: suboptimal tables query schema filter happens late

avallete:

- Filtering on queries happen after the initial gathering happened over all the elements. This PR change it to include the filters (at least the schemaFilter) at the root of the query. Speeding introspections queries up for type generation as long as an `includedSchema` field is present.
- Rework the queries so the `retrieve` also efficiently filter upstream queries

- Tested this to also re-generate the types on `postgrest-js` repo, no diff showed up, which help ensure a bit more that there is no functional changes in the results of the introspection.
- Tested over a "problematic" database with a lot of schemas that timeouted with types introspections, went from a ~22s running time to ~200ms to generate the types.


src/lib/sql/functions.sql.ts
```diff
@@ -21,7 +29,40 @@ with functions as (
...
+                ),
+                ' '
+              ) AS arr
+            ) AS split_args
+          ) args
+    )`
+                : "''"
+            } AND`
+          : ''
+    }
```
avallete, line 65: **todo:**

I'm not sure this is well covered by the tests, might wanna add some to make sure this doesn't break behaviour.

soedirgo (approved):

PR #993, opened 2025-09-24
chore: dedup typescript typegen logic

avallete:

- Follow up refactor isolating duplicated typescript generation logic into their own dedicated functions

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -32,19 +32,12 @@ export const apply = async ({
   postgrestVersion?: string
 }): Promise<string> => {
   schemas.sort((a, b) => a.name.localeCompare(b.name))
-
-  const columnsByTableId = Object.fromEntries<PostgresColumn[]>(
-    [...tables, ...foreignTables, ...views, ...materializedViews].map((t) => [t.id, []])
+  relationships.sort(
+    (a, b) =>
+      a.foreign_key_name.localeCompare(b.foreign_key_name) ||
+      a.referenced_relation.localeCompare(b.referenced_relation) ||
+      JSON.stringify(a.referenced_columns).localeCompare(JSON.stringify(b.referenced_columns))
   )
-  for (const column of columns) {
-    if (column.table_id in columnsByTableId) {
-      columnsByTableId[column.table_id].push(column)
-    }
-  }
-  for (const tableId in columnsByTableId) {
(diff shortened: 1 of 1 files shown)


src/server/templates/typescript.ts
```diff
@@ -628,7 +630,7 @@ export const Constants = {
             )}
           }
         }`
-    })}
+  })}
```
avallete, line 633: **note**

Forgot to remove this loop, now we can use the introspectionSchema directly where the values are already filtered and sorted.

src/server/templates/typescript.ts
```diff
@@ -147,117 +312,68 @@ export type Database = {
...
                     })}
                   }
                   Relationships: [
-                    ${relationships
-                      .filter(
-                        (relationship) =>
-                          relationship.schema === table.schema &&
-                          relationship.referenced_schema === table.schema &&
-                          relationship.relation === table.name
-                      )
```
avallete, line 240: **note**

Noticed that I forgot this loop, instead I now push those into the `introspectedSchema` object under each table/view so we don't have to filter and sort everytime.

Also extracted the typegen into it's own dedicated function reducing duplication across tables/views definitions.

src/server/templates/typescript.ts
```diff
@@ -93,32 +161,105 @@ export const apply = async ({
...
-  const typesById = types.reduce(
-    (acc, type) => {
-      acc[type.id] = type
-      return acc
-    },
-    {} as Record<number, (typeof types)[number]>
-  )
+  const getFunctionTsReturnType = (fn: PostgresFunction, returnType: string) => {
+    return `${returnType}${fn.is_set_returning_function ? '[]' : ''}`
+  }
```
soedirgo, line 174: Only used in one place, I think we can inline this

src/server/templates/typescript.ts
```diff
@@ -93,32 +161,105 @@ export const apply = async ({
...
+        schemas,
+        tables,
+        views,
+      })
+    }
+
+    return 'unknown'
+  }
+
+  const getFunctionSignatures = (
```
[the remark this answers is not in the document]
avallete: Same: https://github.com/supabase/postgres-meta/pull/971/files#diff-c903ec71df16a8ae5f665c19c21b273f74c2f8eb02f5e8a3552ebe69d18be18fR405-R473

soedirgo (approved):

PR #1005, opened 2025-10-24
fix(postgres): remove unnecessary schemaFilter in tables query by id

blinkmj:

Bug fix

When retrieving a specific table by ID, the query incorrectly included a schemaFilter which could potentially cause issues with finding the table. The schemaFilter was unnecessarily applied in the context of querying by ID since IDs are unique across schemas.

## What is the new behavior?

Removed the unnecessary schemaFilter when querying tables by ID. This makes the query more reliable and efficient when fetching a specific table by its unique identifier.

## Additional context

This change affects the retrieve method in PostgresMetaTables when fetching a table by its ID. The fix ensures that tables are properly retrieved by ID regardless of their schema, as the ID itself is sufficient to uniquely identify the table.


avallete:
Thanks for notifying and contributing. It has been fixed upstream. Closing this as duplicate.

PR #1025, opened 2025-12-05
feat: add Dart generation capability

leomos:

Feature

Dart types generation is not supported

## What is the new behavior?

Dart types generation is supported

## Additional context

- Test is failing, cannot figure out the problem for the life of me since there is no real and visible `mismatch` between the snapshot and the generated code; maybe some invisible character?
- Not a Dart expert, I expect some patterns might be improved
- Not a TypeScript expert either ( 😄 ) so the generation code is slightly different from the other generators and it works by creating a TypeScript object for each dart type (`DartType` interface) that is going to be eventually declared (`Declarable` interface); this was done mostly to wrap my head around the whole generation logic
- It supports most builtin types, enums, tables, views, composite types and lists of all of those (apart from views)
- It does not support materialized views


spydon:
Thanks for your effort to bring this into the package!
We're going a bit of a different route with this, so it's superseded by supabase/supabase-flutter#1635

PR #1036, opened 2026-01-31
RFC: Add Zod and JSON Schema type generators

kiwicopple:
Adds planning docs and documentation for two new type generators:

- **TypeScript Zod** (`/generators/typescript-zod`) — Generates Zod runtime validation schemas from the database, producing `Row`, `Insert`, and `Update` schemas per table. Lets users validate data at runtime, not just at compile time.
- **JSON Schema** (`/generators/jsonschema`) — Generates a JSON Schema (Draft 2020-12) document. Language-agnostic, can be used for validation in any language or as input to other code generators.

Also adds a `docs/` folder with documentation for **all 6 generators** (TypeScript, Go, Swift, Python, and the two new ones). Each doc includes usage examples, type mappings, and query parameters.

## What's in this PR

This PR is **planning and documentation only** — no implementation code yet.

| File | What it is |
|---|---|
| `plan.md` | Implementation plan with architecture, type mappings, file changes, and implementation order |
| `docs/typescript.md` | Docs for the existing TypeScript generator |
| `docs/go.md` | Docs for the existing Go generator |
| `docs/swift.md` | Docs for the existing Swift generator |
| `docs/python.md` | Docs for the existing Python generator |
| `docs/typescript-zod.md` | Docs for the new Zod generator (planned) |
| `docs/jsonschema.md` | Docs for the new JSON Schema generator (planned) |

## Key design decisions

See `plan.md` for full details.

- **`db_driver_type` parameter** — Both new generators accept `direct` (default) or `postgrest` to handle type differences between direct PG drivers and PostgREST (e.g., timestamps as `Date` vs strings).
- **`{language}-{format}` naming** — The Zod generator is `typescript-zod`, not just `zod`, so generators are grouped by language. Extensible for future variants (`python-pydantic`, `typescript-io-ts`, etc).
- **Schema namespacing** — Zod schemas are nested under per-schema namespace objects to avoid collisions when the same table name exists in multiple schemas.
- **No new dependencies** — Zod output is plain TypeScript (users install `zod` themselves). JSON Schema is plain JSON.
- **Reuses `GeneratorMetadata`** — Zero changes to the core introspection layer.


docs/typescript.md
```diff
@@ -0,0 +1,178 @@
...
+        }
+        Relationships: [...]
+      }
+    }
+    Views: { ... }
+    Functions: { ... }
+    Enums: { ... }
+    CompositeTypes: { ... }
+  }
+}
```
avallete: **note**

Might wanna do it another way instead. Split things down into interfaces first, then reference them into the Database type something like:

```ts
export interface PublicTablesUsersRow {
   ....
}

export interface PublicTablesUsersInsert {
   ...
}

export type Database = {
    public: {
         Tables: {
             Row: PublicTablesUsersRow,
             Insert: PublicTablesUsersInsert,
             ...
         }
    }
}
```

avallete:
Status note (2026-09-07): type generation has moved out of this repo (v0.99.0 / #1084) into `@supabase/postgrest-typegen`. The in-repo generators are frozen pending a move into per-SDK repos, so this RFC is no longer actionable here. Flagging in case it can be closed. cc @lukasklingsbo

PR #1044, opened 2026-02-22
feat(types): add bigint support for PostgreSQL int8 (64-bit integer)

akbarfaa:
This PR extends the original change (bigint support in the TypeScript generator) and adds the required fixes to make the Docker test suite green and stable across environments.

Feature + test stability fixes:

- Feature: add bigint support for 64‑bit integers in the TypeScript generator.
- Stability: make Docker-based tests pass consistently (Windows/Linux) by fixing connection handling, timeouts, and platform‑specific test scripts.
- PostgreSQL int8 (64‑bit signed integer) is mapped to number , which can overflow JS safe integers.
- Docker tests can fail due to:
  - connection header overwrites,
  - inconsistent DB ports,
  - timeout vs parser‑error ordering,
  - SSL expectations in non‑SSL Docker containers,
  - Windows env var handling.
## What is the new behavior?
- int8 now maps to number | bigint (additive and backward‑compatible).
- Docker tests are stable and green:
  - respects existing pg header values,
  - uses port 5433 for the Docker test DB,
  - tolerates timeout vs parser error behavior,
  - SSL tests accept expected errors based on environment,
  - test scripts work on Windows and Linux.
## Tests
- npm test (Docker) ✅
- npm run check ✅
## Additional context
- The bigint change is scoped to int8 only to avoid semantic shifts for int2/int4/float4/float8/numeric .
- CI should now pass the full Docker test matrix.

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -421,7 +421,7 @@ export const apply = async ({
               const type = typesById.get(type_id)
               let tsType = 'unknown'
               if (type) {
-                tsType = pgTypeToTsType(schema, type.name, {
+                tsType = pgTypeToTsRpcArgType(schema, type.name, {
                   types,
                   schemas,
                   tables,
@@ -440,7 +440,7 @@ export const apply = async ({
               const type = typesById.get(type_id)
               let tsType = 'unknown'
               if (type) {
-                tsType = pgTypeToTsType(schema, type.name, {
+                tsType = pgTypeToTsRpcArgType(schema, type.name, {
                   types,
                   schemas,
                   tables,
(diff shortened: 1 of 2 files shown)


avallete (changes requested):
There is a lot of changes that seems unrelated to the issue.

Also even tough the types would say so, I'm unsure if postgrest would actually result into a "bigint" when those kind of columns get queried @steve-chavez does it trim the result down to a js "number" or does it returns a "bigint" ?

Also according to docs: `unlike Number, BigInt cannot represent decimal values and cannot be mixed with Number in operations without explicit conversion` I'm not sure that having both in an union is the solution here. Maybe we can rather have number, or bigint (if postgrest return such values) and feature flag this breaking change on the typegen.

steve-chavez:
> @steve-chavez does it trim the result down to a js "number" or does it returns a "bigint" ?

We return the bigint but IIRC the we truncate those numbers on the postgrest-js side, see https://github.com/supabase/postgrest-js/issues/319#issuecomment-1550157345.

avallete:

PR #1048, opened 2026-02-28
fix: map vector type to number[] in TypeScript type generation

nancysangani:
Bug fix

`vector` type in RPC function parameters generates as `string` in TypeScript. The Supabase client requires `number[]` for vector parameters — `JSON.stringify()` does not work as a workaround for RPC functions and causes queries to fail or return zero results.

Fixes #1029

## What is the new behavior?
`vector` type now correctly generates as `number[]`.

Before:
```typescript
match_documents: {
  Args: {
    query_embedding: string  // ❌
  }
}
```

After:
```typescript
match_documents: {
  Args: {
    query_embedding: number[]  // ✅
  }
}
```

## Additional context
`vector` was grouped with string types in `pgTypeToTsType`. Moved it to its own condition returning `number[]`. A unit test for `pgTypeToTsType` is added in `test/lib/typegen-typescript.test.ts` covering `vector → number[]` and `_vector → (number[])[]`.

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -901,11 +901,12 @@ export const pgTypeToTsType = (
       'timestamp',
       'timestamptz',
       'uuid',
-      'vector',
       'interval',
     ].includes(pgType)
   ) {
     return 'string'
+  } else if (pgType === 'vector') {
+    return 'number[]'
   } else if (['json', 'jsonb'].includes(pgType)) {
     return 'Json'
   } else if (pgType === 'void') {
--- a/test/lib/typegen-typescript.test.ts
+++ b/test/lib/typegen-typescript.test.ts
@@ -1,0 +1,27 @@
+import { expect, test, describe } from 'vitest'
+import { pgTypeToTsType } from '../../src/server/templates/typescript.js'
+
+const mockSchema = { id: 1, name: 'public' } as any
+const mockContext = { types: [], schemas: [], tables: [], views: [] }
+
+describe('pgTypeToTsType', () => {
+  test('maps vector to number[]', () => {
+    expect(pgTypeToTsType(mockSchema, 'vector', mockContext)).toBe('number[]')
+  })
+
+  test('maps _vector to (number[])[]', () => {
+    expect(pgTypeToTsType(mockSchema, '_vector', mockContext)).toBe('(number[])[]')
+  })
+
+  test('maps text to string', () => {
+    expect(pgTypeToTsType(mockSchema, 'text', mockContext)).toBe('string')
+  })
+
+  test('maps bool to boolean', () => {
+    expect(pgTypeToTsType(mockSchema, 'bool', mockContext)).toBe('boolean')
+  })
+
+  test('maps int4 to number', () => {
+    expect(pgTypeToTsType(mockSchema, 'int4', mockContext)).toBe('number')
+  })
+})


avallete (commented):
Thank's for contributing.

Only issue I have with this PR is the global scope of it, and lack of e2e tests about how this will work with `postgrest-js` end to end. I think the next steps we need is:

We need to first verify if this change would:

1. Fix the issue on `postgrest-js` side
2. Have any noticeable side effect

To do so, we need to add more tests to reproduce the initial issue here:

https://github.com/supabase/supabase-js/blob/master/packages/core/postgrest-js/test/supabase/migrations/00000000000000_schema.sql

And add some e2e tests:

https://github.com/supabase/supabase-js/blob/master/packages/core/postgrest-js/test/rpc.test.ts

Also add some tests with the behavior for similar types for insert/update/select on views/table.

Demonstrating that re-generating the types with this change doesn't BC something else and fix the original issue.

We currently don't have a good easy way to do so, that would be the first thing to do.
I think we need to change/add a scripts in `postgrest-js`: 
https://github.com/supabase/supabase-js/blob/master/scripts/generate-postgrest-types.js

So we can easily point it to a local `postgres-meta` repo and regenerate the types from this version.
Right now the only way I see to do so, is to do:

```
cd supabase-js/packages/core/postgrest-js/test
npx supabase start db # start the test database
cd postgres-meta
PG_META_DB_URL="postgresql://postgres:postgres@127.0.0.1:54322/postgres" npm run dev # start pg-meta dev server and link it to the postgrest-js test db
 curl --location 'http://localhost:1337/generators/typescript?included_schemas=public,personal&detect_one_to_one_relationships=true' -o supabase-js/packages/core/postgrest-js/test/types.generated.ts # regenerate the type tests for postgrest-js
cd supabase-js/packages/core/postgrest-js && npm run test:run # run the tests
```

With this new context we can safely proceed to deploy this more globally.

nancysangani:
> Thank's for contributing.
> 
> Only issue I have with this PR is the global scope of it, and lack of e2e tests about how this will work with `postgrest-js` end to end. I think the next steps we need is:
> 
> We need to first verify if this change would:
> 
> 1. Fix the issue on `postgrest-js` side
> 2. Have any noticeable side effect
> 
> To do so, we need to add more tests to reproduce the initial issue here:
> 
> https://github.com/supabase/supabase-js/blob/master/packages/core/postgrest-js/test/supabase/migrations/00000000000000_schema.sql
> 
> And add some e2e tests:
> 
> https://github.com/supabase/supabase-js/blob/master/packages/core/postgrest-js/test/rpc.test.ts
> 
> Also add some tests with the behavior for similar types for insert/update/select on views/table.
> 
> Demonstrating that re-generating the types with this change doesn't BC something else and fix the original issue.
> 
> We currently don't have a good easy way to do so, that would be the first thing to do. I think we need to change/add a scripts in `postgrest-js`: https://github.com/supabase/supabase-js/blob/master/scripts/generate-postgrest-types.js
> 
> So we can easily point it to a local `postgres-meta` repo and regenerate the types from this version. Right now the only way I see to do so, is to do:
> 
> ```
> cd supabase-js/packages/core/postgrest-js/test
> npx supabase start db # start the test database
> cd postgres-meta
> PG_META_DB_URL="postgresql://postgres:postgres@127.0.0.1:54322/postgres" npm run dev # start pg-meta dev server and link it to the postgrest-js test db
>  curl --location 'http://localhost:1337/generators/typescript?included_schemas=public,personal&detect_one_to_one_relationships=true' -o supabase-js/packages/core/postgrest-js/test/types.generated.ts # regenerate the type tests for postgrest-js
> cd supabase-js/packages/core/postgrest-js && npm run test:run # run the tests
> ```
> 
> With this new context we can safely proceed to deploy this more globally.

Thanks for the feedback. I’ll add tests to reproduce the issue, extend e2e coverage, and verify there are no breaking changes with regenerated types. I’ll also look into improving the type generation workflow.

avallete:

PR #1057, opened 2026-04-05
fix: handle empty segments in formatForGoTypeName

### LINKED ISSUE #1056: Go type generator crashes on names starting with underscore
`formatForGoTypeName` in `src/server/templates/go.ts` crashes with a TypeError when processing names that start with a non-alphanumeric character (e.g. `_prisma_migrations`, `_internal`).

The function splits on non-alphanumeric characters, which produces an empty string as the first array element. Then `""[0].toUpperCase()` throws because `""[0]` is `undefined`.

The Python equivalent (`formatForPyClassName`) already handles this with an `if (word)` guard. The Swift equivalent also handles it. The Go version is the only one missing the check.

Common tables affected: `_prisma_migrations` (Prisma), `_internal`, `_metadata`, any underscore-prefixed identifiers.

oniani1:
Fixes #1056

`formatForGoTypeName` crashes with a TypeError on names starting with non-alphanumeric characters (like `_prisma_migrations`). The split produces an empty string element, and indexing into it gives `undefined`.

Added the same `if (word)` guard that `formatForPyClassName` in `python.ts` already uses.

### DIFF
--- a/src/server/templates/go.ts
+++ b/src/server/templates/go.ts
@@ -102,7 +102,13 @@ ${compositeTypes
 function formatForGoTypeName(name: string): string {
   return name
     .split(/[^a-zA-Z0-9]/)
-    .map((word) => `${word[0].toUpperCase()}${word.slice(1)}`)
+    .map((word) => {
+      if (word) {
+        return `${word[0].toUpperCase()}${word.slice(1)}`
+      } else {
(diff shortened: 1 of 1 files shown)


avallete (approved):

avallete:

PR #1062, opened 2026-04-11
fix: generate Insert/Update types for views with INSTEAD OF triggers

### LINKED ISSUE #850: [supabase gen] complex Views with triggers missing Insert/Update definitions
# Bug report

## Describe the bug

`supabase gen types` does not properly handle the case that a complex view uses triggers to write information back through the view. It does not produce the `Insert`/ `Update` sections of the type definitions. When using simple "automatically updatable view this works as expected.

## To Reproduce

We have two tables:
```
CREATE TABLE IF NOT EXISTS private.profile_type (
    id int2 NOT NULL PRIMARY KEY,
    name text NOT NULL,
);
```

```
CREATE TABLE IF NOT EXISTS private.profile (
  id uuid REFERENCES auth.users NOT NULL primary key,
  username text unique,
  profile_type_id int2 REFERENCES private.profile_type(id),
);
```

This view provides access to the profile and integrates information from the other table:
```
CREATE OR REPLACE VIEW public."Profile" ("id", "username", "profileType")
WITH (security_invoker) 
AS SELECT p.id, username, pt.name
FROM private.profile AS p
JOIN private.profile_type AS pt ON p.profile_type_id = pt.id;
```

This view is not an "automatically updatable view". In order to write back to the profile table, we are attaching a trigger to the view that handle the INSERT and UPDATE statements:

```
CREATE OR REPLACE FUNCTION public.profile_view_v1_row()
RETURNS TRIGGER
AS $$
DECLARE
  v_gender_type_id int2;
BEGIN

  IF TG_OP = 'INSERT' THEN
    -- omitted code
  ELSE -- 'UPDATE'
    -- omitted code
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql SECURITY DEFINER;

CREATE OR REPLACE TRIGGER profile_view_v1_row
  INSTEAD OF INSERT OR UPDATE ON public."Profile"
  FOR EACH ROW
  EXECUTE FUNCTION public.profile_view_v1_row();
```

When generating the types using 

```
supabase gen types --lang typescript
```

it does not include the sections for Insert and Update in the public->Views->Profile part of the generated types (see below)

## Expected behavior

The expected type definition should look like this:
```
...
  public: {
    Tables: {
      [_ in never]: never
    }
    Views: {
      Profile: {
        Row: {
          id: string | null
          profileType: string | null
          username: string | null
        }

--> These parts are missing and should be generated
        Insert: {
          id?: string | null
          profileType?: string | null
          username?: string | null
        }
        Update: {
          id?: string | null
          profileType?: string | null
          username?: string | null
        }
<--
...
}
```

## System information

- OS: macOS
- supabase cli version (`supabase --version`): 1.223.10

nancysangani:
Bug fix

Views that use `INSTEAD OF INSERT OR UPDATE` triggers do not get `Insert`/`Update` type definitions generated by `supabase gen types`. Only simple auto-updatable views produce these types.

Fixes #850

## What is the new behavior?
Views with `INSTEAD OF INSERT` or `INSTEAD OF UPDATE` triggers now correctly get `Insert` and `Update` type definitions generated. The two are handled independently, so a view with only an INSERT trigger gets only an `Insert` type, and vice versa.

## Additional context
Root cause: the generator was gating `Insert`/`Update` emission solely on `pg_relation_is_updatable()`, which returns `false` for any complex view (joins, expressions) regardless of whether `INSTEAD OF` triggers make it writable.

Fix: extended the views SQL query to also check `pg_trigger` for `INSTEAD OF` triggers, exposing `is_insert_enabled` and `is_update_enabled` fields separately from `is_updatable`.

### DIFF
--- a/src/lib/sql/views.sql.ts
+++ b/src/lib/sql/views.sql.ts
@@ -11,6 +11,50 @@ SELECT
   c.relname AS name,
   -- See definition of information_schema.views
   (pg_relation_is_updatable(c.oid, false) & 20) = 20 AS is_updatable,
+  -- A view supports INSERT if it is auto-updatable OR has an INSTEAD OF INSERT trigger
+  (
+    (pg_relation_is_updatable(c.oid, false) & 8) = 8
+    OR EXISTS (
+      SELECT 1 FROM pg_trigger t
+      WHERE t.tgrelid = c.oid
+        AND t.tgtype & 64 > 0
+        AND t.tgtype & 4 > 0
+        AND NOT t.tgisinternal
+    )
+  ) AS is_insert_enabled,
+  -- A view supports UPDATE if it is auto-updatable OR has an INSTEAD OF UPDATE trigger
+  (
+    (pg_relation_is_updatable(c.oid, false) & 4) = 4
+    OR EXISTS (
+      SELECT 1 FROM pg_trigger t
+      WHERE t.tgrelid = c.oid
+        AND t.tgtype & 64 > 0
+        AND t.tgtype & 16 > 0
+        AND NOT t.tgisinternal
+    )
+  ) AS is_update_enabled,
+  -- A view supports INSERT if it is auto-updatable OR has an INSTEAD OF INSERT trigger
+  (
+    (pg_relation_is_updatable(c.oid, false) & 8) = 8
+    OR EXISTS (
+      SELECT 1 FROM pg_trigger t
+      WHERE t.tgrelid = c.oid
+        AND t.tgtype & (1 << 2) > 0  -- INSTEAD OF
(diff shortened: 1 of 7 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. The problem you identified (views made writable by INSTEAD OF triggers getting no Insert or Update types) is fixed in supabase/sdk#125 with credit to this PR, implemented via `pg_relation_is_updatable(oid, true)` with independent Insert and Update gating so trigger-writable columns are covered too.

PR #1063, opened 2026-04-13
fix: add | null to composite type fields in TypeScript generator

### LINKED ISSUE #763: Composite types can be null
# Bug report

## Describe the bug

This PR introduced support for composite types, https://github.com/supabase/postgres-meta/pull/459. However, composite type fields can be null.

> The syntax is comparable to CREATE TABLE, except that only field names and types can be specified; no constraints (such as NOT NULL) can presently be included. Note that the AS keyword is essential; without it, the system will think a different kind of CREATE TYPE command is meant, and you will get odd syntax errors.[^1]

Composite type declaration field type values then should output with a `| null`.

An aside: it would also be nice to have support for domains, currently they are generated as unknown type.

```sql
drop type t;

create domain notnullint as int not null;

create type t as (a int, b notnullint);

select (1,2)::t;
-- (1,2)

select (1,null)::t;
-- [2024-04-24 09:54:44] [23502] ERROR: domain notnullint does not allow null values
```

Screenshot 2024-04-24 at 09 57 34

nancysangani:
Bug fix

Composite type fields were not emitting `| null` in the generated TypeScript types, making them incorrectly non-nullable.

Fixes #763

## What is the new behavior?
All composite type fields now emit `| null`, including domain-typed fields which previously fell through to `unknown` without `| null`.

Before:
```ts
CompositeTypes: {
  t: {
    a: number
    b: unknown
  }
}
```

After:
```ts
CompositeTypes: {
  t: {
    a: number | null
    b: unknown | null
  }
}
```

## Additional context
Per the PostgreSQL docs, composite type fields cannot have NOT NULL constraints, so every field is inherently nullable.

### DIFF
--- a/src/server/templates/python.ts
+++ b/src/server/templates/python.ts
@@ -198,6 +198,20 @@ class PythonEnum implements Serializable {
   }
 }
 
+class PythonDomain implements Serializable {
+  name: string
+  py_type: PythonType
+
+  constructor(name: string, schema: string, py_type: PythonType) {
+    this.name = `${formatForPyClassName(schema)}${formatForPyClassName(name)}`
+    this.py_type = py_type
+  }
+
+  serialize(): string {
+    return `${this.name}: TypeAlias = ${this.py_type.serialize()}`
+  }
+}
+
 type PythonType = PythonListType | PythonSimpleType
 
 class PythonSimpleType implements Serializable {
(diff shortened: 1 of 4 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. The underlying issue you identified (composite type attributes are inherently nullable in Postgres) is fixed for the Python generator in supabase/sdk#124 with credit to this PR. The TypeScript side is already handled on current master, where composite attributes emit `| null`.

PR #1064, opened 2026-04-17
Update generate-python-types-test.ts

saiganesh47:

## What is the new behavior?

## Additional context


avallete:

The generators in the new package are also frozen for now — they're being relocated one more time, into each language's own SDK repository (tracked internally). Once the Python generator lands in its SDK repo, that's the place for this change. Closing here as the target code no longer exists.

PR #1072, opened 2026-04-29
feat: apply frozen=True to all basemodels in python generated models

va1ha11a:
Feature
When generating models to python the update and insert typed dicts make sense to be mutable but the base models do not (to me at least) and immutability provides some extra protections. i.e. foot gun removal.

Pydantic basemodels are produced mutably 

Issue is here
https://github.com/supabase/cli/issues/5142

@o-santi 

## What is the new behavior?
Pydantic models are set to frozen


avallete:

The generators in the new package are also frozen for now — they're being relocated one more time, into each language's own SDK repository (tracked internally). Once the Python generator lands in its SDK repo, that's the place for this change. Closing here as the target code no longer exists.

PR #1074, opened 2026-04-29
fix: forward types in pgTypeToGoType array fallback so enum arrays resolve to []string

### LINKED ISSUE #1073: Go typegen returns []interface{} for arrays of enums and composite types
`pgTypeToGoType` in `src/server/templates/go.ts` resolves arrays via:

```ts
if (pgType.startsWith('_')) {
  const innerType = pgTypeToGoType(pgType.slice(1), nullable)
  return \`[]\${innerType} \`
}
```

The recursive call drops the third argument, so `types` defaults to `[]`. Without the type list, the inner enum or composite lookup at lines 302 and 315 finds nothing and falls through to `interface{}` at line 329. The non-array path resolves the same enum to `string` correctly, only the array path is broken.

Repro: column `tags user_status[]` where `user_status` is a public enum. Go output:

```go
Tags []interface{} \`json:"tags"\`
```

Expected:

```go
Tags []string \`json:"tags"\`
// or for nullable
Tags []*string \`json:"tags"\`
```

Forwarding `types` in the recursive call resolves it.

oniani1:
`pgTypeToGoType` recurses through the array branch with only `(pgType, nullable)`, so the third argument defaults to `[]`. Inner enums and composites then can't be resolved and fall through to `interface{}`, even though the same enum resolves to `string` outside an array.

This forwards `types` to the recursive call.

Added `test/server/templates/go.test.ts` with three cases: non-nullable array of enum should be `[]string`, nullable array of enum should be `[]*string`, and plain `text[]` continues to be `[]string`. Calls `apply` with mock `GeneratorMetadata` so the test runs without Docker.

Closes #1073

### DIFF
--- a/src/server/templates/go.ts
+++ b/src/server/templates/go.ts
@@ -321,7 +321,7 @@ function pgTypeToGoType(pgType: string, nullable: boolean, types: PostgresType[]
 
   // Arrays
   if (pgType.startsWith('_')) {
-    const innerType = pgTypeToGoType(pgType.slice(1), nullable)
+    const innerType = pgTypeToGoType(pgType.slice(1), nullable, types)
     return `[]${innerType} `
   }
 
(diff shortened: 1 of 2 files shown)


avallete (approved):

PR #1075, opened 2026-05-04
chore: use literal for limits and offsets

staaldraad:

Bug fix

Limits and offsets are typed as Number but are treated as string in SQL template generation

## What is the new behavior?

Treats these as literals so the SQL will always be correct / valid.

## Additional context

### DIFF
--- a/src/lib/sql/column_privileges.sql.ts
+++ b/src/lib/sql/column_privileges.sql.ts
@@ -1,3 +1,4 @@
+import { literal } from 'pg-format'
 import type { SQLQueryPropsWithSchemaFilter } from './common.js'
 
 export const COLUMN_PRIVILEGES_SQL = (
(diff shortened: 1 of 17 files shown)


avallete (approved):
We do the escaping for `filterByValue` and `filterByList` even tough at the end of the day postgres-meta is meant to execute queries on database that the user already has access to.<br><br>Since the pattern is so similar everywhere, we might consider normalizing this into an helper. That can be done as a follow-up tough.

staaldraad:
> We do the escaping for `filterByValue` and `filterByList` even tough at the end of the day postgres-meta is meant to execute queries on database that the user already has access to.Since the pattern is so similar everywhere, we might consider normalizing this into an helper. That can be done as a follow-up tough.

> postgres-meta is meant to execute queries on database that the user already has access to

💯 agree, there are some endpoints though that have unexpected side effects, for example config.sql.ts can be called as: 
```
http://127.0.0.1/config?limit=1%3BCREATE%20TABLE....
```
Pipeline-ing the `create table...` statement with the config read operation. This just tightens things up a little bit

PR #1077, opened 2026-06-01
fix: type ROWS 1 set-returning functions as arrays

peter941221:
## Problem

TypeScript type generation used `prorows > 1` to decide whether an RPC return type should be an array.

`prorows` comes from PostgreSQL `ROWS n`, which is a planner estimate. It does not change the runtime shape of a set-returning function. A function declared as `RETURNS SETOF ... ROWS 1` still returns a list through PostgREST.

That made generated RPC types collapse `RETURNS SETOF ... ROWS 1` functions to a single object.

## Change

Use `is_set_returning_function` as the return-shape signal for generated TypeScript RPC types.

This keeps scalar and table-returning functions on the existing single-object path, while `RETURNS SETOF ... ROWS 1` now gets an array `Returns` type.

The existing snapshots now reflect the corrected `Returns` and `SetofOptions.isOneToOne` behavior for the `ROWS 1` fixtures. This PR also adds a dedicated regression test that contrasts the two relevant cases directly:

- `function_using_table_returns` stays a single object
- `function_using_setof_rows_one` still generates an array even with `ROWS 1`

## Issue

Addresses supabase/supabase#46525.

## Validation

- `npm run check`
- `npx prettier --check src/server/templates/typescript.ts test/server/typegen.ts`
- `npx -p node@20 node ./node_modules/vitest/vitest.mjs run test/index.test.ts -t "typegen: typescript keeps ROWS 1 set-returning functions as arrays" --reporter verbose`

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -220,7 +220,7 @@ export const apply = async ({
       fn.return_type_id
     )
     const returnsSetOfTable = fn.is_set_returning_function && fn.return_type_relation_id !== null
-    const returnsMultipleRows = fn.prorows !== null && fn.prorows > 1
+    const returnsMultipleRows = fn.is_set_returning_function
     // Case 1: if the function returns a table, we need to add SetofOptions to allow selecting sub fields of the table
     // Those can be used in rpc to select sub fields of a table
     if (returnTableName) {
--- a/test/server/typegen.ts
+++ b/test/server/typegen.ts
@@ -1,6 +1,4 @@
-import CryptoJS from 'crypto-js'
 import { expect, test } from 'vitest'
-import { CRYPTO_KEY } from '../../src/server/constants'
 import { app } from './utils'
 
 test('typegen: typescript', async () => {
@@ -6968,21 +6966,7 @@ test('typegen: python w/ excluded/included schemas', async () => {
 })
 
 test('typegen: typescript keeps ROWS 1 set-returning functions as arrays', async () => {
-  const connectionString = process.env.PG_META_TEST_CONNECTION_STRING
-  const encryptedConnection = connectionString
-    ? CryptoJS.AES.encrypt(connectionString, CRYPTO_KEY).toString()
-    : undefined
-  const { body } = await app.inject({
-    method: 'GET',
-    path: '/generators/typescript',
-    ...(encryptedConnection
-      ? {
-          headers: {
-            'x-connection-encrypted': encryptedConnection,
-          },
-        }
-      : {}),
-  })
+  const { body } = await app.inject({ method: 'GET', path: '/generators/typescript' })
 
   const setofRowsOneMatch = body.match(
     /function_using_setof_rows_one:[\s\S]*?function_using_table_returns:/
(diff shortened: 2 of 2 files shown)


SahilArate:
The one-line fix is correct and well-targeted is_set_returning_function is exactly the right signal here.
One thing worth noting: the PR updates existing snapshot tests but doesn't add a dedicated regression test for the specific scenario from the issue a RETURNS TABLE (...) ROWS 1 function. Without it, a future refactor could silently reintroduce the same regression.
The commenter vaguul on issue #46525 suggested a clean pair of cases worth adding: one scalar function that should type as a single object, and one ROWS 1 set-returning function that should type as []. Adding that to test/server/typegen.ts would make the intent explicit and protect against regression.

peter941221:
Added a dedicated regression test in test/server/typegen.ts for the two cases from the review: function_using_table_returns stays a single object, while function_using_setof_rows_one still generates an array even with ROWS 1. I also kept the assertion focused on the generated function fragments so it locks the return-shape behavior without depending on the full inline snapshot formatting.

avallete (changes requested):
Hey there !

I would like this change backed by an e2e test on the `postgrest-js` repo.
I think this is tested against the actual postgrest runtime:

https://github.com/supabase/supabase-js/blob/a25062e9285fa8e3bd702c59ddda0d87ad1fcc27/packages/core/postgrest-js/test/supabase/migrations/00000000000000_schema.sql#L371

And the behavior seems correct. Embedding result in an ojbect:

https://github.com/supabase/supabase-js/blob/a25062e9285fa8e3bd702c59ddda0d87ad1fcc27/packages/core/postgrest-js/test/embeded_functions_join.test.ts#L382

While direct RPC call result in an array:

https://github.com/supabase/supabase-js/blob/a25062e9285fa8e3bd702c59ddda0d87ad1fcc27/packages/core/postgrest-js/test/advanced_rpc.test.ts#L1231

The type inference job seems to be doing it's job.

peter941221:
I retraced this through the downstream consumer and the patch is changing the contract in the wrong place.

The generated type for function_using_setof_rows_one already keeps Returns as a single row and marks SetofOptions as isOneToOne: true and isSetofReturn: true:
https://github.com/supabase/supabase-js/blob/master/packages/core/postgrest-js/test/types.generated.ts#L754-L765

postgrest-js uses that combination to infer an array for direct rpc calls:
https://github.com/supabase/supabase-js/blob/master/packages/core/postgrest-js/src/types/common/rpc.ts#L126-L131

That matches the existing runtime tests:
direct rpc returns an array
https://github.com/supabase/supabase-js/blob/a25062e9285fa8e3bd702c59ddda0d87ad1fcc27/packages/core/postgrest-js/test/advanced_rpc.test.ts#L1230-L1258

embedded function joins return an object
https://github.com/supabase/supabase-js/blob/a25062e9285fa8e3bd702c59ddda0d87ad1fcc27/packages/core/postgrest-js/test/embeded_functions_join.test.ts#L382-L428

My patch turns Returns into Row[] and flips isOneToOne to false, which removes the direct-rpc vs embedded distinction. I am closing this PR.

PR #1082, opened 2026-06-09
fix: escape identifiers in the Python type generator

### LINKED ISSUE #1081: Python type generator does not escape enum labels and column names
The Python type generator interpolates enum labels and column names into the generated source without escaping.

Enum `Literal[...]` values:

https://github.com/supabase/postgres-meta/blob/f21a4da30484c6146a6ac4a3f260d992e7330bf1/src/server/templates/python.ts#L196

```ts
const variants = this.variants.map((item) => `"${item}"`).join(', ')
```

`Field(alias=...)`:

https://github.com/supabase/postgres-meta/blob/f21a4da30484c6146a6ac4a3f260d992e7330bf1/src/server/templates/python.ts#L240

```ts
return `    ${this.name}: ${py_type} = Field(alias="${this.pg_name}")`
```

A name that contains a double quote, backslash, or newline produces invalid Python and can inject text into the generated module. The TypeScript template already passes every identifier through `JSON.stringify`, so the same input is handled there.

I'll send a PR that wraps these with `JSON.stringify` (which produces a valid Python string literal) and adds a test, matching the TypeScript template.

oniani1:
The Python type generator interpolates enum labels and column names into the generated source without escaping, so a name that contains a double quote, backslash, or newline produces invalid Python and can inject text into the generated module. The TypeScript template already passes every identifier through `JSON.stringify`. This does the same for enum `Literal` values and `Field` aliases.

Added a test that generates types for an enum label and a column name containing a double quote, and checks the output stays inside the string literal. It fails without the change and passes with it.

Closes #1081

### DIFF
--- a/src/server/templates/python.ts
+++ b/src/server/templates/python.ts
@@ -193,7 +193,7 @@ class PythonEnum implements Serializable {
     this.variants = type.enums
   }
   serialize(): string {
-    const variants = this.variants.map((item) => `"${item}"`).join(', ')
+    const variants = this.variants.map((item) => JSON.stringify(item)).join(', ')
     return `${this.name}: TypeAlias = Literal[${variants}]`
   }
 }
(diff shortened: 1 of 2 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. Your fix (escaping identifiers in the Python generator) has been ported in supabase/sdk#124 with credit to this PR.

PR #1084, opened 2026-06-20
feat: consume @supabase/postgrest-typegen for type generation

avallete:
## Summary

Replaces postgres-meta&#39;s embedded type-generation templates/SQL with the released [`@supabase/postgrest-typegen`](https://github.com/supabase/sdk/tree/main/packages/postgrest-typegen) package (the same engine, extracted to `supabase/sdk` and published to npm). `src/lib/generators.ts` becomes a thin adapter that wraps `pgMeta.query` into the package&#39;s structural `Queryable` and calls `introspect()`, preserving the historical `getGeneratorMetadata` signature and `{ data, error }` contract.

Two deliberate behavior changes, both covered by tests:

- Metadata is passed through the package&#39;s `sortGeneratorMetadata` before generation, making output ordering deterministic (semantic sort instead of catalog order); the typegen snapshots were regenerated for this. The generated *content* is unchanged — sorting the lines of `test/server/typegen.ts` before and after leaves only the new default-schema test as a real difference — but consumers regenerating types will see one large reordering diff.
- `getGeneratorMetadata` now ends the connection pool on error paths too, where the previous implementation only ended it on success.

## Status

- The dependency is pinned to the released `^0.2.0` from npm (previously a `pkg.pr.new` preview build, then `^0.1.0`). 0.2.0 formats with oxfmt instead of prettier (typescript snapshots regenerated; the only output change is oxfmt parenthesizing conditional types in generic-default positions, which is semantically inert), runs its introspection queries concurrently (restoring the old CLI-path parallelism), and accepts a `format` hook, which stays unused here since the whole `generateTypescript` call already runs on the worker.
- Merged latest `master`. The template fixes that landed after the extraction are already part of the released package with identical logic:
  - #1096 (only collapse `ROWS 1` return types when `SetofOptions` is emitted)
  - #1101 (Map-based relationship lookup instead of quadratic filtering)
- Review pass caught and fixed a regression: the old typescript template read `PG_META_GENERATE_TYPES_DEFAULT_SCHEMA` directly, so the `/generators/typescript` route honored it; the package takes it as an option, and the route wasn&#39;t passing it. The route now passes `defaultSchema`, with a regression test.

## Worker-thread generation (#1102) preserved

#1102 added opt-in worker-thread formatting (`PG_META_FORMAT_IN_WORKER`) with load shedding. The package formats internally, so formatting alone can no longer be intercepted on this side; instead the whole `generateTypescript` call is handed to the worker. Measured on a synthetic public schema (12 columns and 2 foreign keys per table), 400 tables: wall clock is unchanged vs inline, with the longest main-thread block dropping to ~20ms. Metadata crosses the thread boundary as a structured clone, which is plain JSON and costs nothing measurable.

`format-pool.ts`/`format-worker.js` keep their master names, identifiers (`FormatQueueFullError`, `destroyFormatPool`, `isFormatPoolActive`) and structure, so the only diff in them is the delegation itself: the worker task carries generator metadata into the package&#39;s `generateTypescript` instead of a prettier payload, and the entry point is `generateTypescriptTypes(metadata, options)` instead of `format(code, options)`. Admission control, per-task timeout, idle pool, the 503 shedding path and the `PG_META_FORMAT_*` env vars are all unchanged from master.

The worker task type excludes the package&#39;s `format` callback option (`Omit`): functions cannot cross the structured-clone worker boundary, so the compiler guarantees one is never sent.

## Upstream changes (supabase/sdk#118, released as 0.2.0)

The remaining review findings are addressed in the package rather than here:

- **oxfmt instead of prettier**: prettier was ~90% of generation cost on a 400-table schema, on machine-generated code nobody hand-diffs. supabase/sdk#118 switches the package&#39;s default formatter to `oxfmt` and adds a `format` hook for callers that want to substitute their own.
- **Sequential introspection**: `introspect()` awaited its ten queries one at a time where the old CLI path used `Promise.all`; supabase/sdk#118 runs them concurrently again.
- **Deleted `test/server/templates/go.test.ts`**: the unit cases (enum-array fallback and friends) are already mirrored upstream in the package&#39;s `test/generation/go.test.ts`, alongside broader generation coverage.
- **Forked introspection SQL**: the package&#39;s copies under `src/introspection/sql/` are now the canonical typegen introspection queries, exercised by the package&#39;s own integration tests and a nightly parity job against real postgres-meta; the copies here under `src/lib/sql/` continue to serve the REST API.
- **prettier dedup risk**: moot once the package drops prettier for oxfmt; this repo&#39;s own `prettier` stays for `Parser.ts` (SQL formatting) and dev formatting.

With oxfmt, inline `generateTypescript` barely blocks the event loop at all (measured max block 11ms at 400 tables, 29ms at 1000, 51ms at 2000 — oxfmt&#39;s napi formatting runs off the JS thread), so the worker is belt-and-braces now; removing the machinery entirely is a candidate follow-up after this merges.

## Validation

- `npm run check` (tsc) passes and `prettier --check` passes against the released package.
- The five `format-pool` tests pass against the released 0.2.0, covering worker/inline parity, the CLI opt-out, 503 load shedding and in-flight accounting.
- Generated output for all four languages (typescript, go, swift, python) is identical in content between this branch and current `master` when run against the same database; ordering differs as described above.
- The full test suite does not run green locally on this machine even on `master` (fixture-loading issues in the local Docker environment); the failure set on this branch matches the `master` baseline, and CI is the arbiter for the full suite.


spydon:
@avallete I've fixed up the last things in here and marked it as ready for review. 😃

avallete:
## Review: are #1101 and #1102 preserved?

**#1101 — kept.** `@supabase/postgrest-typegen@0.1.0` carries the fix verbatim: `relationshipsByRelation` built in one pass keyed on `schema.relation`, `getRelationships()` doing an O(1) lookup then filtering on `referenced_schema`. Same code as 923965b, down to the early `if (!candidates) return []`.

**#1102 — not kept.** `routes/generators/typescript.ts` calls `generateTypescript()` directly; the package formats with prettier inline (`src/generation/typescript.ts`, end of `generateTypescript`) and its options are `detectOneToOneRelationships` / `postgrestVersion` / `defaultSchema` only — no format hook. So `format-pool.ts`, `format-worker.js`, the `piscina` dependency, the build step copying the worker into `dist/`, `test/server/format-pool.ts` and the `PG_META_FORMAT_*` docs in CLAUDE.md all survive as dead weight, the 503 load-shedding path is gone, and `PG_META_FORMAT_IN_WORKER=true` silently does nothing. `npm run check` passes, so nothing catches it.

### Where the time actually goes

Measured against the released package on a synthetic public schema (12 columns and 2 foreign keys per table):

| tables | output | `generateTypescript` | prettier share |
|---|---|---|---|
| 100 | 170 KiB | 439 ms | ~73% |
| 400 | 668 KiB | 1065 ms | ~90% |
| 1000 | 1664 KiB | 2504 ms | ~92% |

So a format hook upstream would recover most of it, but not all — the string building ahead of prettier is CPU-bound on the main thread too.

### Suggested fix, working against 0.1.0 today

Hand the whole `generateTypescript` call to the worker instead of just the prettier pass. Same 400-table schema: **1059 ms on a worker vs 1065 ms inline, longest main-thread block down from ~1000 ms to 21 ms.** Metadata crosses the thread boundary as a structured clone, which is plain JSON here and costs nothing measurable.

Pushed to [`claude/postgres-meta-refactor-review-4c4a13`](https://github.com/supabase/postgres-meta/compare/claude/funny-tesla-wofjdj...claude/postgres-meta-refactor-review-4c4a13) (branched off this PR's head). It renames `format-pool`/`format-worker` to `typegen-pool`/`typegen-worker` since they no longer format, keeps the admission control, per-task timeout, idle pool and 503 shedding exactly as they were, and keeps the `PG_META_FORMAT_*` env var names so deployments don't need reconfiguring. The five pool tests are adapted and pass; `tsc` and `prettier --check` are clean. Full suite not run — no Docker in this environment.

### While we're here: prettier is the wrong tool for this

~90% of type generation is prettier, and the output is machine-generated code that nobody diffs by hand. `oxfmt` is roughly an order of magnitude faster on this kind of workload and would cut the cost far more than any threading work. Worth raising against `postgrest-typegen` — it already uses `oxfmt`/`oxlint` on its own source, so the dependency is familiar there. That would need a formatter-choice option (or just a swap, if the output churn is acceptable once), and it makes the worker a nice-to-have rather than a necessity.

### Smaller things

- **CLI type generation went from parallel to sequential.** `server.ts` used to issue the nine introspection queries under `Promise.all`; `introspect()` awaits them one at a time. On a remote database that's 9 × RTT added to every `gen:types:*` run.
- **`test/server/templates/go.test.ts` is deleted** (106 lines) with no replacement in this repo. The go path is still covered by the `typegen.ts` snapshot, but the unit-level cases are gone unless they're mirrored upstream.
- **The introspection SQL is now forked**: `src/lib/sql/*.sql.ts` here and the package's own copies in `src/introspection/sql/`. A fix to one won't reach the other, and only this repo's copy has the test suite pointed at it.
- **"Byte-identical output" in the PR body is contradicted two paragraphs above it.** What's true: sorting the lines of `test/server/typegen.ts` on both branches leaves only the new `PG_META_GENERATE_TYPES_DEFAULT_SCHEMA` test as a real difference — everything else is pure reordering from `sortGeneratorMetadata`. Content identical, order not. Users regenerating types will get one large no-op diff.
- **The CLAUDE.md diff is mostly unrelated prettier reflow** (blank lines before fenced blocks and lists). Worth splitting out so the two real edits are visible.
- `prettier` is currently deduped: the package pins `3.5.3` and this repo's `^3.3.3` resolves to the same. Bumping the root past 3.5.x would quietly nest a second copy and could shift generated formatting.

---
_Generated by [Claude Code](https://claude.ai/code)_

avallete:
Correction to my review above: the #1102 fix is now **in this PR**, not on a separate branch. `4ed0083` is pushed onto `claude/funny-tesla-wofjdj` — disregard the `claude/postgres-meta-refactor-review-4c4a13` link in the previous comment (same commit, now redundant; I could not delete that branch from here, the push was refused).

The PR description is updated to match: the "remaining blocker" section is replaced by what actually landed.

Everything else in the review stands as written — #1101 verbatim in the package, the benchmark numbers, the oxfmt suggestion, and the smaller findings (sequential introspection queries in the CLI path, the deleted go template test, the forked introspection SQL, the CLAUDE.md prettier reflow, the deduped prettier pin). The "byte-identical output" wording is corrected in the description to "identical content, different order".

---
_Generated by [Claude Code](https://claude.ai/code)_

avallete:
Let's hold unti: https://github.com/supabase/sdk/pull/118 get merged

spydon:
@avallete https://github.com/supabase/sdk/pull/118 is merged and postgrest-typegen v0.2.0 is now released and depended on by this PR :)

avallete (commented):
LGTM ! @soedirgo need an approve 🙏

spydon (approved):
It's happening!

PR #1085, opened 2026-06-20
fix(typescript): generate non-nullable json columns as NonNullable<Json>

Maliik-B:
## Problem

A `NOT NULL` `json`/`jsonb` column generates the TypeScript type `Json`, but the generated `Json` type is itself nullable:

```ts
export type Json = string | number | boolean | null | { [key: string]: Json | undefined } | Json[]
```

So the column type structurally permits `null` even though Postgres forbids it. Studio reports the column as non-nullable, but the generated types disagree.

Closes #1055.

## Fix

`generateNullableUnionTsType` already special-cases `unknown`/`any` (which include `null`) by not appending `| null`. This applies the same reasoning to `Json`: a non-nullable `json`/`jsonb` column is narrowed to `NonNullable<Json>` so the type reflects the database constraint.

```ts
if (tsType === 'Json' && !isNullable) {
  return `NonNullable<${tsType}>`
}
```

## Scope

- Only non-nullable `json`/`jsonb` columns change (`Json` becomes `NonNullable<Json>`). Nullable `json` columns are untouched, since `Json` already covers `null`.
- The change sits in the single nullability chokepoint, so it covers Row, Insert, and Update consistently. Function return types and composite-type attributes call this helper with `isNullable: true`, so they are unaffected. Array types (`(Json)[]`) are unaffected.

## Tests

Added a regression test that spins up a throwaway schema with a `NOT NULL jsonb` column and a nullable `jsonb` column, generates types scoped to it, and asserts the non-nullable column is `NonNullable<Json>` while the nullable one stays `Json | null`. It follows the inline-schema pattern from the existing included/excluded-schemas test, so it adds no fixture churn to the shared snapshots.

## Note on the type-generation extraction

I see #1084 and supabase/pg-toolbelt#302 are moving these templates into `@supabase/postgrest-typegen`. This patch targets `typescript.ts` here; happy to port the same one-line guard to the extracted package instead if that is the better home. @avallete

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -483,6 +483,11 @@ export const apply = async ({
     : ''
 
   function generateNullableUnionTsType(tsType: string, isNullable: boolean) {
+    // The Json type already includes `null`, so a NOT NULL json/jsonb column has to be
+    // narrowed with NonNullable to reflect the database constraint. See supabase/postgres-meta#1055.
+    if (tsType === 'Json' && !isNullable) {
+      return `NonNullable<${tsType}>`
+    }
     // Only add the null union if the type is not unknown as unknown already includes null
     if (tsType === 'unknown' || tsType === 'any' || !isNullable) {
       return tsType
--- a/test/server/typegen.ts
+++ b/test/server/typegen.ts
@@ -6964,3 +6964,46 @@ test('typegen: python w/ excluded/included schemas', async () => {
     })
   }
 })
+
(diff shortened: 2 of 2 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. Your fix (narrowing non-nullable json columns to `NonNullable<Json>`) has been ported in supabase/sdk#125 with credit to this PR.

PR #1086, opened 2026-06-22
feat(typegen): add Kotlin type generation

AndroidPoet:
## What

Adds a **Kotlin** target to `gen types`, alongside the existing TypeScript, Go, Swift, and Python generators. It follows the same structure as `swift.ts` (per-schema namespace, `Select`/`Insert`/`Update` variants, enums, composite types) but emits idiomatic `kotlinx.serialization` code.

```bash
PG_META_GENERATE_TYPES=kotlin npm run gen:types:kotlin
# or via the API:
GET /generators/kotlin?included_schemas=public&visibility=internal
```

## Why

The CLI already proxies `supabase gen types --lang {typescript,go,swift,python}` straight through to this service via `PG_META_GENERATE_TYPES`. Kotlin is the one major Supabase client ecosystem with no first-party type generation, despite an active KMP/Android user base. This closes that gap with no new runtime dependencies.

## Example output

```kotlin
import kotlinx.serialization.SerialName
import kotlinx.serialization.Serializable
import kotlinx.serialization.json.JsonElement

object PublicSchema {
  @Serializable
  enum class UserStatus {
    ACTIVE,
    @SerialName("not-valid id") not_valid_id,
  }
  @Serializable
  class EmptyTableSelect
  @Serializable
  data class ProfilesSelect(
    @SerialName("class") val `class`: String,
    @SerialName("created_at") val createdAt: String,
    val id: Long,
    val metadata: JsonElement? = null,
    val status: UserStatus? = null,
    val tags: List<String>? = null,
  )
  // ...Insert / Update variants
}
```

## Design notes

Modeled on `swift.ts`, with a few Kotlin-idiomatic refinements:

- **`@SerialName` only when needed** — emitted only where the camelCased property name differs from the raw column name, instead of an always-present `CodingKeys` block.
- **`= null` defaults on nullable properties** — so `Insert`/`Update` payloads can omit them, matching how Supabase Kotlin clients build partial rows.
- **Column-less relations → plain `@Serializable class`** — a `data class` with an empty primary constructor is a Kotlin compile error, so empty tables/views are special-cased.
- **Keyword safety** — Kotlin hard keywords (`class`, `object`, `in`, …) are backtick-escaped; enum values that aren't valid identifiers are sanitized and preserved via `@SerialName`.
- **Scoped imports** — the import block only includes the symbols actually referenced (`SerialName`, `JsonElement`, `JsonObject`).
- **Type mapping** — `int2/4/8 → Short/Int/Long`, `float4/8 → Float/Double`, `numeric → Double` (no dependency-free KMP-wide decimal; mirrors the Go template), `json/jsonb → JsonElement`, `record → JsonObject`, arrays → `List<T>`, enums/composites → generated types.

Visibility is controlled by `PG_META_GENERATE_TYPES_KOTLIN_VISIBILITY` (`public` | `internal`, default `public`) and the `visibility` query param — analogous to Swift's `access_control`. `public` is emitted implicitly (Kotlin's default) to keep output lint-clean.

## Tests

Added `typegen: kotlin` and `typegen: kotlin w/ internal visibility` to `test/server/typegen.ts`, mirroring the Swift cases. They use empty inline snapshots that populate on first run.

> Note: the snapshot suite requires Docker (per `CLAUDE.md`), which wasn't available in my environment, so the inline snapshots are intentionally left empty for `npm run test:update` to fill against the test DB. `npm run check` (tsc) and `prettier --check` both pass. I verified the generator's actual output against synthetic metadata covering enums, arrays, composite types, nullability, keyword columns, and empty tables (see example above).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,6 +48,7 @@ npm run gen:types:typescript    # Generate TypeScript types from DB schema
 npm run gen:types:python        # Generate Python types (Pydantic models)
 npm run gen:types:go            # Generate Go types
 npm run gen:types:swift         # Generate Swift types (beta)
+npm run gen:types:kotlin        # Generate Kotlin types (kotlinx.serialization)
 
 # With custom DB connection:
 PG_META_DB_URL=postgresql://... npm run gen:types:typescript
@@ -121,7 +122,7 @@ Type generation (`npm run gen:types:*`) works by:
 4. Templates output type definitions to stdout
 
 Environment variables:
-- `PG_META_GENERATE_TYPES`: Language (typescript, python, go, swift)
+- `PG_META_GENERATE_TYPES`: Language (typescript, python, go, swift, kotlin)
 - `PG_META_GENERATE_TYPES_INCLUDED_SCHEMAS`: Comma-separated schemas to include
 - `PG_META_GENERATE_TYPES_DETECT_ONE_TO_ONE_RELATIONSHIPS`: Enable 1:1 relationship detection
 - `PG_META_POSTGREST_VERSION`: PostgREST version for TypeScript template compatibility
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "gen:types:go": "PG_META_GENERATE_TYPES=go node --loader ts-node/esm src/server/server.ts",
     "gen:types:swift": "PG_META_GENERATE_TYPES=swift node --loader ts-node/esm src/server/server.ts",
     "gen:types:python": "PG_META_GENERATE_TYPES=python node --loader ts-node/esm src/server/server.ts",
+    "gen:types:kotlin": "PG_META_GENERATE_TYPES=kotlin node --loader ts-node/esm src/server/server.ts",
     "start": "node dist/server/server.js",
     "dev": "trap 'npm run db:clean' INT && run-s db:clean db:run &&  run-s dev:code",
     "dev:code": "nodemon --exec node --loader ts-node/esm src/server/server.ts | pino-pretty --colorize",
--- a/src/server/constants.ts
+++ b/src/server/constants.ts
@@ -2,6 +2,7 @@ import crypto from 'crypto'
 import { PoolConfig } from '../lib/types.js'
 import { getSecret } from '../lib/secrets.js'
 import { AccessControl } from './templates/swift.js'
+import { Visibility } from './templates/kotlin.js'
 import pkg from '#package.json' with { type: 'json' }
 
 export const PG_META_HOST = process.env.PG_META_HOST || '0.0.0.0'
@@ -50,6 +51,9 @@ export const GENERATE_TYPES_SWIFT_ACCESS_CONTROL = process.env
   .PG_META_GENERATE_TYPES_SWIFT_ACCESS_CONTROL
   ? (process.env.PG_META_GENERATE_TYPES_SWIFT_ACCESS_CONTROL as AccessControl)
   : 'internal'
+export const GENERATE_TYPES_KOTLIN_VISIBILITY = process.env.PG_META_GENERATE_TYPES_KOTLIN_VISIBILITY
+  ? (process.env.PG_META_GENERATE_TYPES_KOTLIN_VISIBILITY as Visibility)
+  : 'public'
 
 // json/jsonb/text types
 export const VALID_UNNAMED_FUNCTION_ARG_TYPES = new Set([114, 3802, 25])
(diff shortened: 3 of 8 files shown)


AndroidPoet:
### End-to-end verification against a live Postgres 15

Ran the actual `PG_META_GENERATE_TYPES=kotlin` path (not just the template in isolation) against a real database — schema with an enum, a composite type, an identity column, a `jsonb` column, an array, a `numeric`, a reserved-word column (`"class"`), an empty table, and a view:

```kotlin
object PublicSchema {
  @Serializable
  enum class UserStatus {
    ACTIVE,
    @SerialName("not-valid id") not_valid_id,
  }
  @Serializable
  class EmptyTableSelect
  @Serializable
  data class ProfilesSelect(
    @SerialName("class") val `class`: String,
    @SerialName("created_at") val createdAt: String,
    @SerialName("home_address") val homeAddress: Address? = null,
    val id: Long,
    val metadata: JsonElement? = null,
    val score: Double,
    val status: UserStatus? = null,
    val tags: List<String>? = null,
  )
  // ProfilesInsert / ProfilesUpdate / ActiveProfilesSelect ...
  @Serializable
  data class Address(
    @SerialName("street_name") val streetName: String,
    val zip: Int,
  )
}
```

The live run caught a bug my isolated test missed: composite attribute types must be resolved by their catalog **`name`** (`int4`), not `format` (`integer`) — otherwise scalars fall through to the `JsonElement` fallback. Fixed in 0320121.

Heads-up for maintainers: the same latent issue exists in `go.ts` and `swift.ts` (both pass `attribute.type.format`); their fixtures just don't include a scalar composite attribute to surface it. Happy to file a separate fix for those if useful.

The two `typegen` snapshot tests still need `npm run test:update` against the Docker test DB to populate their inline snapshots, since the CI fixture differs from my ad-hoc schema.

AndroidPoet:
### Proof the output is real, compilable Kotlin (not just well-formed text)

Compiled the generated file with the actual Kotlin toolchain — `kotlinc` + the bundled `kotlinx-serialization` compiler plugin (runtime 1.11.0, JDK 17):

```
$ kotlinc Generated.kt -Xplugin=kotlinx-serialization-compiler-plugin.jar \
    -cp kotlinx-serialization-core-jvm.jar:kotlinx-serialization-json-jvm.jar -d out.jar
EXIT CODE: 0
```

The `@Serializable` plugin emitted real serializers for every type — confirming the annotations are processed, not decorative:

```
generated/PublicSchema$ProfilesSelect.class
generated/PublicSchema$ProfilesSelect$$serializer.class
generated/PublicSchema$Address$$serializer.class
generated/PublicSchema$UserStatus$Companion.class
... (one $$serializer per data class)
```

Then a full JSON round-trip through `ProfilesSelect`, decoding an actual Postgres-shaped row and re-encoding it:

```kotlin
val row = Json.decodeFromString(PublicSchema.ProfilesSelect.serializer(), json)
```
```
decoded:    id=7 class=`premium` status=ACTIVE tags=[a, b] zip=90210 street=Main St
re-encoded: {"class":"premium","created_at":"2026-06-22T00:00:00Z",
             "home_address":{"street_name":"Main St","zip":90210},
             "id":7,"metadata":{"k":1},"score":9.5,"status":"ACTIVE","tags":["a","b"]}
```

That one test exercises every non-trivial path, and all pass:

- **`@SerialName` mapping** — `created_at`/`home_address` ↔ `createdAt`/`homeAddress`
- **Reserved-word column** — backtick-escaped `` `class` `` decodes and re-encodes
- **Enum** — `status: UserStatus` decodes `"ACTIVE"`
- **Nested composite** — `home_address` → `Address(streetName: String, zip: Int)`
- **Array / jsonb / numeric / int4** — `List<String>` / `JsonElement` / `Double` / `Int` all correct

Full chain verified: live Postgres → pg-meta introspection → `apply()` → valid Kotlin → compiles → serializes round-trip.

YaminMahdi:
Excellent work. Hopefully, it will be merged.

avallete:
Thanks for this. Type generation has moved out of `postgres-meta`: since v0.99.0 this repo consumes [`@supabase/postgrest-typegen`](https://github.com/supabase/sdk/tree/main/packages/postgrest-typegen) and the template layer this PR adds to was deleted in #1084.

New languages are now meant to be built as consumers of the `GeneratorMetadata` JSON contract that package exposes, living next to the SDK they target (the Dart generator in supabase-flutter is the reference for this pattern). A Kotlin generator would follow the same shape, in the Kotlin SDK's repo. Closing here as the target code no longer exists; cc @lukasklingsbo for where a Kotlin generator should live.

PR #1096, opened 2026-08-03
fix(typegen): only collapse ROWS 1 return types when SetofOptions is emitted

### LINKED ISSUE #46525: Type generation incorrectly treats PostgreSQL ROWS 1 RPC functions as returning a single object instead of an array
# Bug report

* [x] I confirm this is a bug with Supabase, not with my own application.
* [x] I confirm I have searched the Docs, GitHub Discussions, and Discord.

## Describe the bug

When generating TypeScript types for PostgreSQL RPC functions, functions declared with `ROWS 1` are incorrectly typed as returning a single object instead of an array of objects.

In PostgreSQL, `ROWS 1` is only a planner hint that indicates the expected number of returned rows. It does **not** enforce a maximum row count, nor does it change the shape of the result returned by PostgREST. PostgREST still returns an array for set-returning functions regardless of the `ROWS` estimate.

However, the generated Supabase types appear to interpret `ROWS 1` as meaning the function returns a single object, causing the generated RPC return type to omit the trailing `[]`.

This leads to a mismatch between the generated types and the actual runtime response shape.

Additionally, calling `.single()` on such an RPC produces confusing typings. The type generation already incorrectly collapses the array due to `ROWS 1`, and `.single()` applies another array-to-object transformation. While the final inferred type may appear correct, it is derived through incorrect assumptions and does not accurately model the underlying API behavior.

## To Reproduce

Create a PostgreSQL function that returns a set and specifies `ROWS 1`:

```sql
create function test_function()
returns table (
  id int,
  name text
)
rows 1
language sql
as $$
  select 1 as id, 'test' as name
$$;
```

Generate Supabase types.

Observe the generated RPC type:

```ts
test_function: {
  Args: Record<PropertyKey, never>
  Returns: {
    id: number
    name: string
  }
}
```

Use the RPC:

```ts
const { data } = await supabase.rpc('test_function')
```

At runtime, PostgREST returns:

```ts
[
  {
    id: 1,
    name: 'test'
  }
]
```

but the generated type indicates:

```ts
{
  id: number
  name: string
}
```

## Expected behavior

Functions declared with `ROWS 1` should still be generated as returning an array type when they are set-returning functions.

Expected generated type:

```ts
test_function: {
  Args: Record<PropertyKey, never>
  Returns: {
    id: number
    name: string
  }[]
}
```

Type generation should reflect the actual response shape returned by PostgREST rather than the PostgreSQL row estimate.

Additionally, `.single()` should be the mechanism that narrows an array response to a single object type, rather than `ROWS 1` influencing the generated return type.

## Screenshots

N/A

## System information

* OS: Any
* Browser (if applicable): Any
* Version of supabase-js: Latest
* Version of Node.js: Any

## Additional context

The issue appears to stem from type generation interpreting PostgreSQL's `ROWS 1` metadata as a guarantee of a single-row result. PostgreSQL documentation describes `ROWS` as an estimate used by the query planner, not a constraint on the number of rows returned.

Because PostgREST continues to serialize set-returning function results as arrays, the generated types become inconsistent with runtime behavior and can cause TypeScript errors or incorrect assumptions in application code.

mandarini:
Fixes supabase/supabase#46525.

The TypeScript typegen drops the `[]` from the return type of every set-returning function declared with `ROWS 1`, but the compensating `SetofOptions` metadata is only emitted when the function returns a named table, view, or composite type. For ad-hoc `RETURNS TABLE (...)` and scalar `SETOF` functions this produced a single-object type while PostgREST still returns an array at runtime, and there is no metadata for postgrest-js to correct it. The collapse is now gated on the same condition as the `SetofOptions` emission, so those functions keep the array type, while named-relation functions (where postgrest-js restores the array for direct RPC calls via `isOneToOne`) are unchanged.

Note for consumers: regenerating types flips `Returns` from object to array for affected functions. The resulting type errors point at real runtime mismatches, since PostgREST has always returned an array for these calls.

### DIFF
--- a/src/server/templates/typescript.ts
+++ b/src/server/templates/typescript.ts
@@ -262,7 +262,12 @@ export const apply = async ({
       }
     }
 
-    return `${returnType}${fn.is_set_returning_function && returnsMultipleRows ? '[]' : ''}
+    // A ROWS 1 estimate may only collapse the array type when SetofOptions is also emitted
+    // (returnTableName resolved): postgrest-js relies on isOneToOne to restore the array shape
+    // for direct RPC calls. Without that metadata PostgREST still returns an array, so the
+    // collapse would make the type diverge from the actual response.
+    const collapsibleSingleRow = !returnsMultipleRows && returnTableName !== null
+    return `${returnType}${fn.is_set_returning_function && !collapsibleSingleRow ? '[]' : ''}
                           ${setofOptionsInfo ? `${setofOptionsInfo}` : ''}`
   }
 
--- a/test/db/00-init.sql
+++ b/test/db/00-init.sql
@@ -501,3 +501,24 @@ STABLE
 AS $$
   SELECT interval_test_row.duration_required * 2;
 $$;
+
+-- Set-returning function with an ad-hoc TABLE return type and a ROWS 1 estimate.
+-- ROWS 1 is only a planner estimate: PostgREST still returns an array, and with no
+-- named relation there is no SetofOptions metadata to restore the array shape, so
+-- the generated return type must keep the array.
+CREATE OR REPLACE FUNCTION public.get_todos_summary_rows_one()
+RETURNS TABLE (id int, name text)
+LANGUAGE SQL STABLE
+ROWS 1
+AS $$
+  SELECT 1 AS id, 'test' AS name;
+$$;
+
+-- Set-returning scalar function with a ROWS 1 estimate: array type must be kept here too.
+CREATE OR REPLACE FUNCTION public.get_todo_ids_rows_one()
+RETURNS SETOF bigint
+LANGUAGE SQL STABLE
+ROWS 1
+AS $$
+  SELECT id FROM public.todos LIMIT 1;
+$$;
--- a/test/server/typegen.ts
+++ b/test/server/typegen.ts
@@ -753,6 +753,7 @@ test('typegen: typescript', async () => {
                   isSetofReturn: true
                 }
               }
+          get_todo_ids_rows_one: { Args: never; Returns: number[] }
           get_todos_by_matview: {
             Args: { "": unknown }
             Returns: {
@@ -841,6 +842,13 @@ test('typegen: typescript', async () => {
                   isSetofReturn: true
                 }
               }
+          get_todos_summary_rows_one: {
+            Args: never
+            Returns: {
+              id: number
+              name: string
+            }[]
+          }
           get_user_audit_setof_single_row: {
             Args: { user_row: Database["public"]["Tables"]["users"]["Row"] }
             Returns: {
@@ -1978,6 +1986,7 @@ test('typegen w/ one-to-one relationships', async () => {
                   isSetofReturn: true
                 }
               }
+          get_todo_ids_rows_one: { Args: never; Returns: number[] }
           get_todos_by_matview: {
             Args: { "": unknown }
             Returns: {
@@ -2066,6 +2075,13 @@ test('typegen w/ one-to-one relationships', async () => {
                   isSetofReturn: true
                 }
               }
+          get_todos_summary_rows_one: {
+            Args: never
+            Returns: {
+              id: number
+              name: string
+            }[]
+          }
           get_user_audit_setof_single_row: {
             Args: { user_row: Database["public"]["Tables"]["users"]["Row"] }
             Returns: {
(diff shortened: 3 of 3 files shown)


mandarini:
## E2E verification against postgrest-js

I ran the cross-repo check locally: built two pg-meta images — one from this PR's merge-base (e7d86a3) and one from this branch — pinned each into the postgrest-js test stack, regenerated types for the postgrest-js test database with both, and compared against PostgREST's actual runtime responses.

### 1. No change to postgrest-js's existing type surface

`supabase gen types typescript --local --schema public,personal` output is **byte-identical** between the merge-base image and this PR's image, for the entire postgrest-js test database.

This is expected: all three `ROWS 1` functions in that schema (`get_user_profile_non_nullable`, `get_user_first_message`, `function_using_setof_rows_one`) return **named relations**, so they keep the collapsed object `Returns` + `SetofOptions.isOneToOne: true`. The contract postgrest-js relies on (`isOneToOne` restoring the array shape for direct RPC calls) is untouched — no inference regression is possible against the current test surface.

### 2. For the broken shapes, the new types match runtime; the old ones don't

Added the affected shapes to the live test DB and called them through PostgREST:

| Function | merge-base typegen | this PR | actual PostgREST response |
|---|---|---|---|
| `RETURNS TABLE(id int, name text) ROWS 1` | `{ id, name }` ❌ | `{ id, name }[]` ✅ | `[{"id":1,"name":"hello_one"}]` |
| `RETURNS SETOF text ROWS 1` | `string` ❌ | `string[]` ✅ | `["hello_scalar"]` |

Neither shape emits `SetofOptions`, so postgrest-js has no metadata to correct the collapsed type — the merge-base types diverge from the actual response, which is exactly supabase/supabase#46525.

### 3. postgrest-js type suite (tstyche)

- Committed `types.generated.ts` (control): **29/29 files pass**
- Fully regenerated types (either image — outputs are identical): 28/29, the one failure being the `scalar_computed_count` / `scalar_computed_ids` assertions in `embeded_functions_join.test.ts`

That failure is **pre-existing drift, not this PR**: it reproduces identically with the unmodified merge-base image. The committed `types.generated.ts` in postgrest-js contains `SetofOptions.isNotNullable` entries (introduced via supabase/supabase-js#2224) that aren't emitted by any pg-meta master build, so regenerating with any current master-based image drops them. Worth tracking separately from this fix.

<details>
<summary>Reproduction steps</summary>

```bash
# 1. Build both images so the Supabase CLI resolves them locally (checked before pulling)
git -C postgres-meta worktree add /tmp/pgmeta-base $(git -C postgres-meta merge-base HEAD master)
docker build postgres-meta -t public.ecr.aws/supabase/postgres-meta:rows1-pr
docker build /tmp/pgmeta-base -t public.ecr.aws/supabase/postgres-meta:rows1-base

# 2. In supabase-js/packages/core/postgrest-js: pin the image tag and start the stack
printf 'rows1-base' > test/supabase/.temp/pgmeta-version
pnpm exec supabase --workdir test start

# 3. Generate with each image (the pin is re-read per invocation)
pnpm exec supabase --workdir test gen types typescript --local --schema public,personal > /tmp/types-base.ts
printf 'rows1-pr' > test/supabase/.temp/pgmeta-version
pnpm exec supabase --workdir test gen types typescript --local --schema public,personal > /tmp/types-pr.ts
diff /tmp/types-base.ts /tmp/types-pr.ts   # → identical

# 4. Runtime source of truth for the affected shapes
docker exec supabase_db_test psql -U postgres -c "
CREATE FUNCTION public.e2e_rows1_ad_hoc_table(prefix text)
RETURNS TABLE(id int, name text) LANGUAGE SQL STABLE ROWS 1
AS \$\$ SELECT 1, prefix || '_one' \$\$;
CREATE FUNCTION public.e2e_rows1_scalar_setof(prefix text)
RETURNS SETOF text LANGUAGE SQL STABLE ROWS 1
AS \$\$ SELECT prefix || '_scalar' \$\$;
NOTIFY pgrst, 'reload schema';"
curl -s http://127.0.0.1:54321/rest/v1/rpc/e2e_rows1_ad_hoc_table \
  -X POST -H "apikey: $ANON" -H "Content-Type: application/json" -d '{"prefix":"hello"}'
# → [{"id":1,"name":"hello_one"}]

# 5. Type suite: swap regenerated types in, post-process, run tstyche
cp /tmp/types-pr.ts test/types.generated.ts
node scripts/update-json-type.js && pnpm exec prettier --write test/types.generated.ts
pnpm exec tstyche
```

</details>

Follow-up regression tests for the ROWS 1 array contract on the SDK side are tracked in SDK-1383.

avallete (approved):
Nice. I think we'll want the new tests to be added to postgrest-js against actual e2e postgrest runtime when this is merged and released. That way we don't get a regression on the behavior in the future 👍

PR #1104, opened 2026-08-14
fix(typegen): omit generated columns from Insert and Update types

### LINKED ISSUE #838: Remove generated columns from Insert/Update types
# Bug report

## Describe the bug

When creating a generated column, it is treated as a regular column in the Studio UI but also when generating TS types. It should be removed from Insert/Update types.

## To Reproduce

1. Create table with generated column:
```sql
CREATE TABLE people (
    ...,
    height_cm numeric,
    height_in numeric GENERATED ALWAYS AS (height_cm / 2.54) STORED
);
```
2. Setup a CLI and generate types.

## Expected behavior

Generated columns do not appear in Insert/Update types.

daviddallakyan2005:
Fixes #838

Stored `GENERATED ALWAYS AS … STORED` columns are not writable, but TypeScript typegen only excluded identity-ALWAYS columns. `is_generated` is already populated by column introspection; Insert/Update now treat it the same way and emit `field?: never`, while the column remains on Row. Go, Swift, and Python generators are separate and still include generated columns — this PR is TypeScript-only.

## How to test

```bash
npm install
npm test
```

`npm test` starts Postgres via `test/db/docker-compose.yml` (Docker required) and runs Vitest. Use Node 22 (`.nvmrc`). The new assertion is `typegen: typescript omits generated columns from Insert and Update` against fixture table `public.people`.


daviddallakyan2005:
Closing to replace with a single-commit PR.

PR #1110, opened 2026-08-18
feat: add json output exposing the GeneratorMetadata contract

spydon:

Feature, stacked on #1084 (base branch is its `claude/funny-tesla-wofjdj` head; only the last commit is new). Adds a `json` output alongside typescript, go, swift, and python, emitting the raw `GeneratorMetadata` contract of `@supabase/postgrest-typegen` that #1084 makes postgres-meta consume.

## What is the new behavior?

- `GET /generators/json` returns the introspected `GeneratorMetadata` (schemas, tables, foreign tables, views, materialized views, columns, relationships, functions, types) as pretty-printed JSON, with the same `included_schemas`/`excluded_schemas` query parameters as the other generators.
- `PG_META_GENERATE_TYPES=json` (and the `gen:types:json` npm script) emits the same document on stdout, so the Supabase CLI can expose it as `supabase gen types --lang json` and pipe it into out-of-process generators as `--lang dart` (supabase/cli#6230, supabase/supabase-flutter#1635 whose parser consumes exactly this contract).

This gives third-party and community type generators a single high-fidelity, language-neutral introspection source: exact nullability, database defaults, identity and generated columns, enum values, views, and materialized views, which lossy sources like the PostgREST OpenAPI description cannot provide.

## Additional context

- The output is a plain `JSON.stringify` over the `sortGeneratorMetadata`-canonicalized introspection result, so it adds no new queries and stays in sync with whatever the built-in generators see.
- Supersedes #1106 (closed as a duplicate of the #1084 extraction): same endpoint surface, now backed by the package instead of the deleted in-repo metadata assembly. Unlike #1106 this emits the bare contract without a `version` envelope; whether the serialized contract should carry one is raised on the postgrest-typegen RFC.
- Tests assert the envelope keys and the fidelity data that motivate the endpoint (nullability, defaults, identity, enum values, views, materialized views), plus schema filtering. The full suite passes locally with the same pre-existing environment-specific failures as `master` (SSL and query-timeout tests).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ Helpers:
   - [ ] GET `/typescript`: Generate Typescript types
   - [ ] GET `/swift`: Generate Swift types (beta)
   - [ ] GET `/python`: Generate Python types (beta)
+  - [ ] GET `/json`: Dump the raw generator metadata as JSON, for third-party type generators
 
 ## Quickstart
 
@@ -110,6 +111,7 @@ where `<lang>` is one of:
 - `go`
 - `swift` (beta)
 - `python` (beta)
+- `json` (the raw generator metadata all the other generators consume, for building your own generator)
 
 To use your own database connection string instead of the provided test database, run:
 `PG_META_DB_URL=postgresql://postgres:postgres@localhost:5432/postgres npm run gen:types:<lang>`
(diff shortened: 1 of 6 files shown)


spydon:
Replaced by the new postgrest-typegen package

PR #1127, opened 2026-08-30
fix(typegen): escape Go struct tags

### LINKED ISSUE #1125: Go type generation emits invalid struct tags for backtick column names
## Bug

PostgreSQL quoted identifiers can contain backticks. The Go type generator currently embeds column names directly in a raw struct-tag literal, so a column such as `bad`tag` terminates the literal early and produces Go source that `gofmt` cannot parse.

## Reproduction

Generate Go types for a table containing a column whose name includes a backtick, then pipe the generated source to `gofmt`. It fails with syntax errors including `expected semicolon` and `string literal not terminated`.

## Expected behavior

The generated struct tag should preserve the PostgreSQL column name while remaining valid Go source. Ordinary column names should retain their current output.

hsusul:
## Summary

- preserve PostgreSQL column and composite-attribute names in valid Go struct tags
- keep the existing raw-tag output for ordinary names and use an interpreted literal only when a backtick requires it
- add focused regression coverage for table columns, composite attributes, and ordinary tags

## Problem

PostgreSQL quoted identifiers can contain backticks. The Go generator interpolated column names directly into raw struct-tag literals, so a name containing a backtick terminated the literal and produced source that `gofmt` rejected.

## Root cause

Both table-column and composite-attribute paths emitted `json` tags without escaping the database-provided name or accounting for the raw literal delimiter.

## Solution

Build the `json` tag value with JSON-compatible quoting. Continue using the current raw Go literal when possible, and emit an interpreted Go literal when the value contains a backtick.

## Tests and validation

- `npx vitest run test/server/templates/go.test.ts` — 6 passed
- generated backtick-name fixture piped to `gofmt` — passed
- `npm run check` — passed
- `npm run build` — passed
- `npx prettier --check src/server/templates/go.ts test/server/templates/go.test.ts` — passed
- `git diff --check` — passed
- Docker-backed suite — 199 passed, 2 unrelated failures on this Node 24 host; the SSL message mismatch also fails on untouched `origin/master`, while the oversized-result timeout passed on the untouched baseline

## Compatibility and risk

Ordinary generated tags retain their previous representation. The change is limited to tag serialization and does not alter field naming or Go type mapping.

## Documentation

No documentation change is needed because this restores valid generated output for supported PostgreSQL identifiers.

Fixes #1125

### DIFF
--- a/src/server/templates/go.ts
+++ b/src/server/templates/go.ts
@@ -112,6 +112,11 @@ function formatForGoTypeName(name: string): string {
     .join('')
 }
 
+function formatForGoStructTag(name: string): string {
+  const tag = `json:${JSON.stringify(name)}`
+  return name.includes('`') ? JSON.stringify(tag) : `\`${tag}\``
+}
+
 function generateTableStruct(
   schema: PostgresSchema,
   table: PostgresTable | PostgresView | PostgresMaterializedView,
@@ -157,7 +162,7 @@ function generateTableStruct(
   const formattedColumnEntries = columnEntries.map(([formattedName, type, name]) => {
     return `  ${formattedName.padEnd(maxFormattedNameLength)} ${type.padEnd(
       maxTypeLength
-    )} \`json:"${name}"\``
+    )} ${formatForGoStructTag(name)}`
   })
 
   return `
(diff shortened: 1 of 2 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. Your fix has been ported in supabase/sdk#122 with credit to this PR; your `JSON.stringify` escaping approach with the interpreted-literal fallback for backticks is the one that was adopted.

PR #1128, opened 2026-08-30
fix(typegen): escape Swift string literals

### LINKED ISSUE #1126: Swift type generation emits invalid string literals for quoted database names
## Bug

PostgreSQL identifiers and enum labels can contain double quotes, backslashes, and control characters. The Swift type generator currently interpolates enum raw values and CodingKeys raw values directly into quoted Swift literals. For example, a value containing a double quote terminates the literal early, and `swiftc -parse` rejects the generated source.

## Reproduction

Generate Swift types for an enum label or column name containing a double quote and pipe the result to `swiftc -parse -`. The parser reports an unterminated string literal and invalid declarations.

## Expected behavior

Generated Swift string literals should escape database-provided values while preserving the original raw enum label or coding key. Ordinary values should retain their current output.

hsusul:
## Summary

- escape database-provided enum raw values and CodingKeys raw values as valid Swift string literals
- cover quotes, backslashes, newlines, and interpolation markers
- add focused regression tests and compiler-parse validation

## Problem

PostgreSQL identifiers and enum labels can contain characters that are special inside Swift strings. The Swift generator interpolated those values directly, allowing quotes to terminate literals and backslash-parenthesis sequences to become interpolation.

## Root cause

The shared enum serializer emitted raw values between quotes without escaping. That serializer is used for both PostgreSQL enums and generated CodingKeys.

## Solution

Add a local Swift string-literal serializer that escapes quotes, backslashes, tabs, line endings, control characters, and Unicode line separators, then use it at the shared enum serialization seam.

## Tests and validation

- `npx vitest run test/server/templates/swift.test.ts` — 3 passed
- generated quote/interpolation fixture piped to `swiftc -parse -` — passed
- `npm run check` — passed
- `npm run build` — passed
- `npx prettier --check src/server/templates/swift.ts test/server/templates/swift.test.ts` — passed
- `git diff --check` — passed
- Docker-backed suite — 199 passed; 2 unrelated failures and 1 skip occurred in the changed run. Untouched `origin/master` reproduced the Node 24 SSL-message failure, while the trigger snapshot and result-size setup passed there, indicating environmental flakiness outside the changed generator path

## Compatibility and risk

Ordinary values serialize identically. Escaping changes only source representation; decoded enum and CodingKeys raw values remain the original PostgreSQL strings.

## Documentation

No documentation change is needed because this restores valid generated output for supported PostgreSQL values.

Fixes #1126

### DIFF
--- a/src/server/templates/swift.ts
+++ b/src/server/templates/swift.ts
@@ -153,6 +153,36 @@ function generateProtocolConformances(protocols: string[]): string {
   return protocols.length === 0 ? '' : `: ${protocols.join(', ')}`
 }
 
+function swiftStringLiteral(value: string): string {
+  let escaped = ''
+  for (const character of value) {
+    const codePoint = character.codePointAt(0)!
+    switch (character) {
+      case '\\':
+        escaped += '\\\\'
+        break
+      case '"':
+        escaped += '\\"'
+        break
+      case '\t':
+        escaped += '\\t'
+        break
+      case '\n':
+        escaped += '\\n'
+        break
+      case '\r':
+        escaped += '\\r'
+        break
+      default:
+        escaped +=
+          codePoint < 0x20 || codePoint === 0x7f || codePoint === 0x2028 || codePoint === 0x2029
+            ? `\\u{${codePoint.toString(16)}}`
+            : character
+    }
+  }
+  return `"${escaped}"`
+}
+
 function generateEnum(
   enum_: SwiftEnum,
   { accessControl, level }: SwiftGeneratorOptions & { level: number }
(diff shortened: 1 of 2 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. Your fix has been ported in supabase/sdk#123 with credit to this PR; your character-loop escaping (including the control character and U+2028/U+2029 handling and the `swiftc -parse` validation) is the approach that was adopted.

PR #1129, opened 2026-08-31
fix(typegen): accept deserialized json/jsonb values in generated types for Python

h-zahar:
The issue was reported in [supabase/supabase-py#1597](https://github.com/supabase/supabase-py/issues/1597).

Currently, the Python generator maps `json` and `jsonb` columns to pydantic's `Json[Any]`, which expects a `str` containing JSON and deserializes it. PostgREST returns those columns already deserialized, so every generated model with a JSON column raises on `model_validate`: JSON input should be string, bytes or bytearray.

With this PR, both column types map to `pydantic.JsonValue`, which denotes an already-deserialized JSON value. 

The issue is resolved using pydantic's native type without defining a custom type.

### DIFF
--- a/src/server/templates/python.ts
+++ b/src/server/templates/python.ts
@@ -46,7 +46,7 @@ from typing import (
     TypedDict,
 )
 
-from pydantic import BaseModel, Field, Json
+from pydantic import BaseModel, Field, JsonValue
 
 ${concatLines(Object.values(ctx.user_enums))}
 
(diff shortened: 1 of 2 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. Your fix (mapping json/jsonb to pydantic's `JsonValue` so deserialized PostgREST payloads validate) has been ported in supabase/sdk#124 with credit to this PR.

PR #1131, opened 2026-08-31
fix: escape backticks in generated Go struct tags

### LINKED ISSUE #1125: Go type generation emits invalid struct tags for backtick column names

Kjubikstronk:
Closes #1125.

The Go generator writes struct tags as a raw string literal, ```go
`json:"NAME"`
```
and Go raw literals cannot contain a backtick at all. A column named `bad`tag` therefore closes the literal early and the generated file is not parseable Go, so `gofmt` reports "string literal not terminated".

Go allows a struct tag to be an interpreted literal too, so this switches to that form only when the tag contains a backtick, escaping backslashes and double quotes. Names without a backtick keep the exact raw-literal output they had before, which is what the issue asks for.

There were two identical call sites, one for tables and views and one for composite types. Both now go through the same helper.

Three tests added to the existing `test/server/templates/go.test.ts`, which is a pure unit test file needing no database: a backtick name produces the escaped interpreted literal, a name with a double quote and a backtick escapes both, and an ordinary name still produces the unchanged raw literal. The first two fail without the change and the third passes either way, so the previous output is pinned as well.

`prettier --check` and `tsc -p tsconfig.json --noEmit` are clean.

### DIFF
--- a/src/server/templates/go.ts
+++ b/src/server/templates/go.ts
@@ -112,6 +112,16 @@ function formatForGoTypeName(name: string): string {
     .join('')
 }
 
+// Go raw string literals cannot contain a backtick, so a column name with one
+// has to be emitted as an interpreted literal instead.
+function formatGoStructTag(name: string): string {
+  const tag = `json:"${name}"`
+  if (!tag.includes('`')) {
+    return `\`${tag}\``
+  }
+  return `"${tag.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`
+}
+
 function generateTableStruct(
   schema: PostgresSchema,
   table: PostgresTable | PostgresView | PostgresMaterializedView,
@@ -157,7 +167,7 @@ function generateTableStruct(
   const formattedColumnEntries = columnEntries.map(([formattedName, type, name]) => {
     return `  ${formattedName.padEnd(maxFormattedNameLength)} ${type.padEnd(
       maxTypeLength
-    )} \`json:"${name}"\``
+    )} ${formatGoStructTag(name)}`
   })
 
   return `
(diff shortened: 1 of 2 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. The struct tag escaping issue this PR addresses is fixed in supabase/sdk#122, which also escapes quotes, backslashes and control characters inside the tag value so `reflect.StructTag.Get` round-trips the exact column name.

Kjubikstronk:
Makes sense, and sdk#122 is the better fix, since escaping quotes, backslashes and control characters means `reflect.StructTag.Get` round-trips anything rather than just surviving a backtick. Closing this one. I had not spotted #1084, so thanks for the pointer.

Kjubikstronk:
Thanks for explaining the move, and for re-landing it rather than just closing it. Your version in supabase/sdk#122 is the better fix: escaping quotes, backslashes and control characters as well means `reflect.StructTag.Get` round-trips any column name, where mine only stopped the backtick from terminating the literal.

One small ask. supabase/sdk#122 links #1125, the issue, but not this PR, so there is no trail from the merged fix back to where the change came from. Would you mind adding a line to the description pointing at it? A "supersedes supabase/postgres-meta#1131" would do it.

Not a complaint about the close, that was clearly the right call once the templates are going away. It just makes the work findable.

PR #1132, opened 2026-08-31
fix: escape database values in generated Swift string literals

### LINKED ISSUE #1126: Swift type generation emits invalid string literals for quoted database names

Kjubikstronk:
Closes #1126.

`generateEnum` interpolates database values straight into a quoted Swift literal:

```ts
case ${case_.formattedName} = "${case_.rawValue}"
```

An enum label or column name containing a double quote closes the literal early, and `swiftc -parse` then rejects the file as an unterminated string literal. Backslashes and control characters cause the same class of problem.

This escapes the value before interpolating: backslash and double quote get their Swift escapes, tab, newline and carriage return use the short forms, and any remaining control character becomes ``u{n}``. Swift unescapes these at compile time, so the raw enum label and coding key still round-trip to the original database value. Values with none of these characters are emitted exactly as before.

That one function generates both the enum cases and the CodingKeys, so both paths mentioned in the issue are covered by the single change.

Added `test/server/templates/swift.test.ts`, a pure unit test file needing no database, in the same style as the existing `go.test.ts`. It covers a quote in an enum label, a quote in a column name, a backslash, and an ordinary name. The first three fail without the change and the last passes either way, so the existing output is pinned too.

`prettier --check` and `tsc -p tsconfig.json --noEmit` are clean.

### DIFF
--- a/src/server/templates/swift.ts
+++ b/src/server/templates/swift.ts
@@ -153,6 +153,27 @@ function generateProtocolConformances(protocols: string[]): string {
   return protocols.length === 0 ? '' : `: ${protocols.join(', ')}`
 }
 
+// Enum labels and column names come from the database and can contain characters
+// that end a Swift string literal early, so escape them before interpolating.
+function escapeForSwiftStringLiteral(value: string): string {
+  return value.replace(/["\\\u0000-\u001f\u007f]/g, (char) => {
+    switch (char) {
+      case '\\':
+        return '\\\\'
+      case '"':
+        return '\\"'
+      case '\n':
+        return '\\n'
+      case '\r':
+        return '\\r'
+      case '\t':
+        return '\\t'
+      default:
+        return `\\u{${char.codePointAt(0)!.toString(16)}}`
+    }
+  })
+}
+
 function generateEnum(
   enum_: SwiftEnum,
   { accessControl, level }: SwiftGeneratorOptions & { level: number }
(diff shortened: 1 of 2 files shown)


spydon:
Thank you for the contribution! postgres-meta's type generation is moving to the shared `@supabase/postgrest-typegen` package in supabase/sdk (see #1084), so open template fixes are being re-landed there. The Swift string literal escaping issue this PR addresses is fixed in supabase/sdk#123, which additionally escapes control characters and U+2028/U+2029.

Kjubikstronk:
Good, sdk#123 is the right home for it. Closing this one.

One thing in case it is useful over there: in this template both the enum cases and the CodingKeys come out of the same `generateEnum`, so a single escape covers both paths. Worth a quick check that the shared package has not split them.

Kjubikstronk:
Same as on #1131, thanks for the explanation and for porting it instead of dropping it. supabase/sdk#123 goes further than mine did, since escaping the C0 controls and U+2028/U+2029 covers cases my version would still have emitted uncompilable Swift for.

Same small ask too: supabase/sdk#123 links #1126 but not this PR, so would you mind adding a "supersedes supabase/postgres-meta#1132" to the description? Purely so the merged fix points back at where it came from.

